PR #9, opened 2019-03-29
Single module project

### LINKED ISSUE #6: Make threadpoolctl a single module project
I think we should remove the package folder structure and just keep everything in a `threadpoolctl.py` module.

We need to make sure to prefix all classes, constants, helper functions with a leading `_` prefix to make it explicit that they are not part of the public API.

I think it's fine not to have the test file as part of the PyPI shipped wheel package.

This would make it easier to vendor.

ogrisel:
Fixes #6 to make it easier to vendor.


ogrisel:
I did the PR from an old master. Let me fix this.

threadpoolctl.py
```diff
@@ -1,16 +1,23 @@
...
+# https://github.com/IntelPython/smp Copyright (c) 2017, Intel Corporation
+# also published under the BSD 3-Clause license
 import os
 import re
 import sys
 import ctypes
 from ctypes.util import find_library
 
-from .utils import _format_docstring
+__version__ = '1.0.0.dev0'
```
ogrisel: Let's not be afraid to start with 1.0.0 for the first release :)

tomMoral:
LGTM! Thanks for the refactoring @ogrisel. Let's merge this.

ogrisel:
Argl, I just noticed that the `tests/` folder is included in the wheel. Let me try to fix that in another PR.

PR #10, opened 2019-03-29
Do not include tests in wheel package

ogrisel:
setuptools is too broken, let's try with flint.


ogrisel:
This is weird, it includes commits that have already been merged in master...

ogrisel:
My bad, I pushed the wrong branch.

PR #12, opened 2019-03-30
[MRG] Refactoring to introduce threadpool_info

### LINKED ISSUE #8: Remove filename_prefixes from the dicts returned by get_threadpool_limits
I don't think the list of `filename_prefixes` is a useful info for the users introspecting their runtime. If they need this they can read the source code of threadpoolctl.

Also maybe `get_threadpool_limits` should be renamed to `get_threadpool_info`.

ogrisel:
This PR fixes #8.

I have also renamed "n_threads" to "num_threads" and "module_path" to "filepath" for consistency.

### DIFF
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -22,7 +22,8 @@ def test_threadpool_limits_by_prefix(openblas_present, mkl_present, prefix):
                             for info in original_infos}
     mkl_found = any([True for info in original_infos
                      if info["prefix"] in ('mkl_rt', 'libmkl_rt')])
-    prefix_found = len([info["prefix"] for info in original_infos])
+    prefix_found = len([info["prefix"] for info in original_infos
+                        if info["prefix"] == prefix])
     if not prefix_found:
         if "mkl_rt" in prefix and mkl_present and not mkl_found:
             raise RuntimeError("Could not load the MKL prefix")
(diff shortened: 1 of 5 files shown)


ogrisel:
I have also refactored the tests to only use the public API so as to make them easier to maintain.

tests/test_threadpoolctl.py
```diff
@@ -106,40 +112,6 @@ def test_set_threadpool_limits_bad_input():
         threadpool_limits(limits=(1, 2, 3))
 
 
-@pytest.mark.parametrize("user_api", (None, "blas", "openmp"))
-def test_thread_limit_context(user_api):
```
tomMoral: It would be nice to test `threadpool_limits` without the context at some point, as it can be used as a function.

ogrisel: Done in 5ec2a71.

tests/test_threadpoolctl.py
```diff
@@ -101,6 +101,26 @@ def test_set_threadpool_limits_by_api(user_api):
     assert threadpool_info() == original_infos
 
 
+def test_threadpool_limits_function_with_side_effect():
+    # Check that threadpool_limits can be used as a function with
+    # side effects instead of a context manager.
+    original_infos = threadpool_info()
+
+    threadpool_limits(limits=1)
```
tomMoral: You could also use `limitation = threadpool_limits(limits=1)` and then use `limitation.unregister()` in the finally to avoid side effects.

ogrisel: Actually, do we really want to make `unregister` public?

ogrisel: I added a test for this in a6acd86.

PR #13, opened 2019-03-31
CI improvements

ogrisel:
Small improvements, will merge if green.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,3 +7,15 @@ module = "threadpoolctl"
 author = "Thomas Moreau"
 author-email = "thomas.moreau.2010@gmail.com"
 home-page = "https://github.com/joblib/threadpoolctl"
+description-file = "README.md"
+classifiers = [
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -7,3 +7,15 @@ module = "threadpoolctl"
 author = "Thomas Moreau"
 author-email = "thomas.moreau.2010@gmail.com"
 home-page = "https://github.com/joblib/threadpoolctl"
+description-file = "README.md"
+classifiers = [
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: BSD License",
+    "Programming Language :: Python :: 2",
+    "Programming Language :: Python :: 2.7",
```
tomMoral: I guess we should maybe remove the support of 2.7?
It does not install correctly if I am not mistaken (see #21 ).

ogrisel: Indeed.

PR #22, opened 2019-06-11
Set env vars to work around multiple OpenMP runtimes bug

jeremiedbb:
Add a `environ_name` field for the supported apis which contains the corresponding env var to control the number of threads:
`MKL_NUM_THREADS`, `OPENBLAS_NUM_THREADS` or `OMP_NUM_THREADS`.

Set those in addition to using the omp_set function. This is a work around for an OpenMP bug(?) when doing nested parallelism with different OpenMP runtimes. With the omp_set functions, only the number of threads in the master thread of the outer loop would be impacted.

### DIFF
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -197,23 +197,26 @@ def test_openmp_nesting(nthreads_outer):
         outer_num_threads, inner_num_threads = \
             check_nested_openmp_loops(10, nthreads)
 
+        # The number of threads available in the outer loop should not have
+        # been decreased. Inner should have been set to 1.
+        assert outer_num_threads == nthreads
+        assert inner_num_threads == 1
+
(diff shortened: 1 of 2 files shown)


jeremiedbb:
doesn't work :(
the env var needs to have been set before the dynamic lib has been loaded.

tomMoral:
This seems weird.
It seems it is no always the case as it is working with some builds but not other no? Have you seen the case with other loops than `gcc` outer and `clang` inner?

jeremiedbb:
> It seems it is no always the case as it is working with some builds but not other no?

I think that when it works, it's not thanks to the setting of the env vars. For example, the test passes on my laptop because inner openmp self restricts to sequential with or without the env vars.

> Have you seen the case with other loops than gcc outer and clang inner?

gcc as inner always restricts itself to sequential.
clang and icc behave the same.

PR #23, opened 2019-06-28
[MRG] ENH Add BLIS support

### LINKED ISSUE #15: Add support for BLIS
This would involve setting up a new CI environment.

The function names are:
- `bli_thread_get_num_threads` / `bli_thread_set_num_threads`
- there is a python package with compiled extension on PyPI: https://pypi.org/project/blis/

 I would be even better to have a CI setup with numpy/scipy built against blis for instance and check that we can control the specific BLIS instance used by those.

jeremiedbb:
Fixes #15

By default, BLIS is sequential and `num_threads` will be -1 in the infos.
It can be a bit confusing since -1 means all cores sometimes, e.g. sklearn. I think we should map -1 to +1. wdyt ?

### DIFF
--- a/continuous_integration/install_with_blis.sh
+++ b/continuous_integration/install_with_blis.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+set -e
+
+pushd ..
+ABS_PATH=$(pwd)
+popd
+
+# Assume Ubuntu: install a recent version of clang and libomp
+echo "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main" | sudo tee -a /etc/apt/sources.list.d/llvm.list
+echo "deb-src http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main" | sudo tee -a /etc/apt/sources.list.d/llvm.list
+wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+sudo apt update
+sudo apt install clang-8 libomp-8-dev
+
+# create conda env
+conda create -n $VIRTUALENV -q --yes python=$VERSION_PYTHON pip cython
+source activate $VIRTUALENV
+
+pushd ..
+
+# build & install blis
+mkdir BLIS_install
+git clone https://github.com/flame/blis.git
+pushd blis
+./configure --prefix=$ABS_PATH/BLIS_install --enable-cblas --enable-threading=openmp CC=$BLIS_CC auto
+make -j4
+make install
+popd
+
+# build & install numpy
+git clone https://github.com/numpy/numpy.git
+pushd numpy
+echo "[blis]
+libraries = blis
+library_dirs = $ABS_PATH/BLIS_install/lib
+include_dirs = $ABS_PATH/BLIS_install/include/blis
+runtime_library_dirs = $ABS_PATH/BLIS_install/lib" > site.cfg
+python setup.py build_ext -i
+pip install -e .
+popd
+
+popd
+
+python -m pip install -q -r dev-requirements.txt
+CFLAGS=-I$ABS_PATH/BLIS_install/include/blis LDFLAGS=-L$ABS_PATH/BLIS_install/lib \
+    bash ./continuous_integration/build_test_ext.sh
+
+python --version
+python -c "import numpy; print('numpy %s' % numpy.__version__)" || echo "no numpy"
+python -c "import scipy; print('scipy %s' % scipy.__version__)" || echo "no scipy"
+
+python -m flit install --symlink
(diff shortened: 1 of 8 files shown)


jeremiedbb:
still WIP because not tested in CI. I'll try to setup some builds with numpy linked against BLIS.

continuous_integration/install_with_blis.sh
```diff
@@ -0,0 +1,52 @@
...
+
+set -e
+CURRENT_PATH=$(pwd)
+
+# Assume Ubuntu: install a recent version of clang and libomp
+# echo "deb http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main" | sudo tee -a /etc/apt/sources.list.d/llvm.list
+# echo "deb-src http://apt.llvm.org/xenial/ llvm-toolchain-xenial-8 main" | sudo tee -a /etc/apt/sources.list.d/llvm.list
+# wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+# sudo apt update
+# sudo apt install clang-8 libomp-8-dev
```
ogrisel: Is this still required? If not we can remove this comment.

jeremiedbb: still wip, I'm updating test_nested_prange_blas to also test with BLIS.

I disabled clang temporarily but I'd like to do 2 jobs: one with clang gcc and another one with gcc clang.

jeremiedbb: Also, by default, BLIS is sequential and num_threads will be -1 in the infos.
It can be a bit confusing since -1 means all cores sometimes, e.g. sklearn. I think we should map -1 to +1. wdyt ?

jeremiedbb:
The uncovered line is the handling of single-threaded blis. In tests I enabled multi-threaded blis.

The timeout job is unrelated to blis. It comes from mkl-clang bad interaction.

ogrisel:
> The timeout job is unrelated to blis. It comes from mkl-clang bad interaction.

The problem is that apparently only this PR can trigger the issue. So maybe it's related to BLIS indirectly somehow?

jeremiedbb:
In addition to the global methods based on environment variables and runtime function calls, BLIS also offers a **local, per-call method of requesting parallelism at runtime**. This method has the benefit of being thread-safe and flexible; your application can spawn two threads at the application level, with each thread requesting different degrees of parallelism from their respective calls to level-3 BLIS operations.
:blush:

**Note: Neither way (automatic nor manual) of specifying multithreading via the local runtime API can be used via the BLAS interfaces.**
:unamused:

tests/_openmp_test_helper/nested_prange_blas.pyx
```diff
@@ -25,18 +41,25 @@ def check_nested_prange_blas(double[:, ::1] A, double[:, ::1] B, int nthreads):
...
-    threadpool_infos = None
+    threadpool_infos = [None]
 
-    for i in prange(n_chunks, num_threads=nthreads, nogil=True):
-        dgemm(trans, no_trans, &n, &chunk_size, &k,
-              &alpha, &B[0, 0], &k, &A[i * chunk_size, 0], &k,
-              &beta, &C[i * chunk_size, 0], &n)
+    with nogil, parallel(num_threads=nthreads):
+        if openmp.omp_get_thread_num() == 0:
+            with gil:
```
jeremiedbb: I figured out what was the cause of the hanging. Note exactly the cause but at least which line of the code was faulty. It happens if we try to call `threadpool_info` in another thread than the main thread (I was doing that in a previous commit).

I don't know why but it's probably related to the bad state of OpenMP threadpool in non main thread when there are multiple openmp loaded.

Anyway, it's unrelated to this PR and should be investigated separately.

ogrisel: I agree. I see that you opened #27 to track this issue.

threadpoolctl.py
```diff
@@ -210,6 +218,10 @@ def threadpool_info():
     modules = _load_modules(user_api=_ALL_USER_APIS)
     for module in modules:
         module['num_threads'] = module['get_num_threads']()
+        # by default BLIS is single-threaded and get_num_threads returns -1.
+        # we map it to 1 for consistency with other libraries.
+        if module['num_threads'] == -1 and module['internal_api'] == 'blis':
+            module['num_threads'] = 1
```
ogrisel: This line is allegedly not covered by any test according to codecov. Do you think it would be possible to make sure it's covered? I believe the only way is to introduce a new build configuration without  `export BLIS_NUM_THREADS=4`.

jeremiedbb:
According to @glemaitre the coverage issue is a false positive.
Maybe due to the fact that I modified a bunch of non python files.

PR #28, opened 2019-09-06
Try to use conda-forge compilers on macOS

ogrisel:
Try to see if `conda install -c conda-forge compilers` works on macOS.


ogrisel:
Fixed in #29 by @jeremiedbb.

PR #38, opened 2019-09-23
[WIP] Make it possible to set limits from already loaded modules

jeremiedbb:
Use case is the following:
```python
infos = threadpool_info(return_api=True)

# create a python threadpool e.g. with joblib threading backend
joblib.Parallel(backend='threading', n_jobs=88)(
    with threadpool_limits(limit, user_api, infos=infos):
        <some function>
)
```

Proposed api:
- add a parameter to `threadpool_info` to return the `set_num_threads` and `get_num_threads` functions
- add a parameter to `threadpool_limits` to accept already loaded modules instead of finding them again.

todo:
- [ ] lots of tests :)
- [ ] update docstrings


jeremiedbb:
The use-case is not achievable. Closing.

ogrisel:
This was implemented in #95 although for another use case.

PR #39, opened 2019-09-24
[MRG] install llvm-openmp in macOS conda-forge job again

jeremiedbb:
Used to be shipped with conda forge compilers. Seems it's not anymore.


ogrisel:
Thanks for the fix.

PR #42, opened 2019-10-21
[MRG] UNIX: dlopen with RTLD_NOLOAD flag

jeremiedbb:
RTLD_NOLOAD is defined in the os module only on unix.
the `mode` parameter is ignored on windows, so it should not matter which value we use for mode. By safety I use the default mode defined by ctypes.

ctypes always add rtld_now, not configurable, to the `mode`. So `mode=RTLD_NOLOAD` should do
`dlopen("path", RTLD_NOW | RTLD_NOLOAD)`.

### DIFF
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -332,15 +332,6 @@ def _match_module(module_info, prefix, prefixes, user_api):
                                    module_info['user_api'] in user_api)
 
 
-def _get_rtld_noload():
-    """Return the RTLD_NOLOAD flags when the OS supports it"""
-    try:
-        rtld_noload = os.RTLD_NOLOAD
-    except AttributeError:
-        rtld_noload = None
-    return rtld_noload
-
-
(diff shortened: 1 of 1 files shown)


jeremiedbb:
Not sure if it's useful in any manner because dlopen does not load again a library if it's already loaded (only increments a ref counter). But I guess it can't hurt either.

tomMoral:
Aht is the rational behind adding this flag? If it is not already loaded, it don't load it?

In this case, maybe also adding `RTLD_NODELETE` would help?

jeremiedbb:
> Aht is the rational behind adding this flag? If it is not already loaded, it don't load it?

Yes, if it's not in the address space, dlopen will return a NULL pointer instead of a valid handle. As far as I understand it's used to check if a shared lib is already loaded or not.

> In this case, maybe also adding RTLD_NODELETE would help?

I don't think so because shared libraries are never closed before the end of the program in python.

Issue #45, opened 2019-10-22
CI random failure when building Blis: Cannot continue with unregistered configuration 'skx'.


ogrisel:
```
configure: detected Linux kernel version 4.15.0-1059-azure.
configure: python interpeter search list is: python python3 python2.
configure: using 'python' python interpreter.
configure: found python version 3.7.4 (maj: 3, min: 7, rev: 4).
configure: python 3.7.4 appears to be supported.
configure: C compiler search list is: gcc clang cc.
configure: using 'gcc' C compiler.
configure: C++ compiler search list is: g++ clang++ c++.
configure: using 'g++' C++ compiler (for sandbox only).
configure: found gcc version 5.4.0 (maj: 5, min: 4, rev: 0).
configure: checking for blacklisted configurations due to gcc 5.4.0.
configure: warning: gcc 5.4.0 does not support 'skx'; adding to blacklist.
configure: checking gcc 5.4.0 against known consequential version ranges.
configure: note: found gcc version older than 6.1.
configure: found assembler ('as') version 2.26.1 (maj: 2, min: 26, rev: 1).
configure: checking for blacklisted configurations due to as 2.26.1.
configure: configuration blacklist:
configure:   skx
configure: reading configuration registry...done.
configure: determining default version string.
configure: found '.git' directory; assuming git clone.
configure: executing: git describe --tags.
configure: got back 0.6.0-58-ge6f0a96c.
configure: truncating to 0.6.0-58.
configure: starting configuration of BLIS 0.6.0-58.
configure: configuring with official version string.
configure: found shared library .so version '3.0.0'.
configure:   .so major version: 3
configure:   .so minor.build version: 0.0
configure: automatic configuration requested.
configure: hardware detection driver returned 'skx'.
configure: checking configuration against contents of 'config_registry'.
configure: 'auto-detected configuration 'skx' is NOT registered!
configure: 
configure: *** Cannot continue with unregistered configuration 'skx'. ***
configure: 
##[error]Bash exited with code '1'.
##[section]Finishing: Install with BLIS
```

ogrisel:
This only happens if the Azure build VM is running on a host with a Intel Skylake X architecture.

mehsanib:
I get similar failure when building configure blis on mac: "auto-detected configuration ' '  "; Did you find out how to fix it? Thanks

ogrisel:
As linked about: https://github.com/joblib/threadpoolctl/pull/46

seuwins:
[...]

PR #46, opened 2019-10-22
CI use recent gcc version to build Blis under Linux

ogrisel:
This should fix: #45.


ogrisel:
Using apt to install gcc-8 did the trick.

PR #47, opened 2019-10-23
[WIP] ENH: add TBB support

### LINKED ISSUE #43: Add support for TBB introspection
It's possible to get a CI env with Threading Building Blocks as follows. At the moment threadpoolctl cannot detect TBB loaded by MKL itself loaded by numpy when `MKL_THREADING_LAYER=TBB` is set:

```
$ conda create -n numpy-tbb numpy tbb
$ conda activate numpy-tbb
$ MKL_THREADING_LAYER=TBB python -c "import numpy; from pprint import pprint; from threadpoolctl import threadpool_info; pprint(threadpool_info())"
[{'filepath': '/home/ogrisel/miniconda3/envs/mkl-threading-layer/lib/libmkl_rt.so',
  'internal_api': 'mkl',
  'num_threads': 2,
  'prefix': 'libmkl_rt',
  'user_api': 'blas',
  'version': '2019.0.4'}]

```

jeremiedbb:
fixes #43 

TBB's api can't be accessed with ctypes (c++ class), so we need to access it through tbb4py.

todo:
- [x] decide what to do when tbb4py is not installed but `libtbb` appears in the module infos. 

- [ ] add tests either through mkl threading layer or directly with tbb threadpools.

### DIFF
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -319,8 +319,7 @@ def _get_version(dynlib, internal_api):
     elif internal_api == "blis":
         return _get_blis_version(dynlib)
     elif internal_api == "tbb":
-        # tbb does not expose it's version
-        return None
+        return _get_tbb_version(dynlib)
     else:
         raise NotImplementedError("Unsupported API {}".format(internal_api))
 
@@ -356,6 +355,11 @@ def _get_blis_version(blis_dynlib):
     get_version = getattr(blis_dynlib, "bli_info_get_version_str")
     get_version.restype = ctypes.c_char_p
     return get_version().decode('utf-8')
+
+def _get_tbb_version(tbb_dynlib):
+    """Return the TBB version"""
+    get_version = getattr(tbb_dynlib, "TBB_runtime_interface_version")
+    return str(get_version())
 
 
 def _tbb_get_num_threads():
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -368,16 +368,16 @@ def _make_module_info(filepath, module_info, prefix):
                     tbb.global_control.max_allowed_parallelism)
     else:
         set_func = getattr(dynlib,
-                        _MAP_API_TO_FUNC[internal_api]['set_num_threads'],
-                        lambda num_threads: None)
+                           _MAP_API_TO_FUNC[internal_api]['set_num_threads'],
+                           lambda num_threads: None)
         get_func = getattr(dynlib,
-                        _MAP_API_TO_FUNC[internal_api]['get_num_threads'],
-                        lambda: None)
+                           _MAP_API_TO_FUNC[internal_api]['get_num_threads'],
+                           lambda: None)
 
     module_info = module_info.copy()
     module_info.update(dynlib=dynlib, filepath=filepath, prefix=prefix,
-                    set_num_threads=set_func, get_num_threads=get_func,
-                    version=_get_version(dynlib, internal_api))
+                       set_num_threads=set_func, get_num_threads=get_func,
+                       version=_get_version(dynlib, internal_api))
     return module_info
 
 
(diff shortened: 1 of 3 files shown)


jeremiedbb:
for the first todo point, there are 2 things

- currently calling threadpool_info when tbb4py is not installed but libtbb is loaded will display `'num_threads: None'`. I would not show None but something like "not available (install tbb4py)". Is it too verbose for this field ?

- currently calling threadpool_limits(1, 'tbb') will silently do nothing if tbb4py is not installed. I think it's not nice. I would raise a warning. Do you agree ?

ogrisel:
> currently calling threadpool_info when tbb4py is not installed but libtbb is loaded will display 'num_threads: None'. I would not show None but something like "not available (install tbb4py)". Is it too verbose for this field ?

I think it's fine to return `num_threads: None`. Maybe we could have a new "comment", "cause" or "status" field in the dict we return that would  "not available (install tbb4py)" in case we cannot instrospect the number of threads.

> currently calling threadpool_limits(1, 'tbb') will silently do nothing if tbb4py is not installed. I think it's not nice. I would raise a warning. Do you agree ?

We should at least warn, possibly raise `ImportError` or `RuntimeError` to tell the user to install tbb4py but only when calling `threadpool_limits` on tbb explicitly.

threadpoolctl.py
```diff
@@ -262,6 +276,9 @@ def _get_version(dynlib, internal_api):
         return _get_openblas_version(dynlib)
     elif internal_api == "blis":
         return _get_blis_version(dynlib)
+    elif internal_api == "tbb":
+        # tbb does not expose it's version
```
anton-malakhov: hmm.. that's an interesting question. adding an issue

tomMoral:
>     * currently calling threadpool_info when tbb4py is not installed but libtbb is loaded will display `'num_threads: None'`. I would not show None but something like "not available (install tbb4py)". Is it too verbose for this field ?

I agree with @ogrisel that it is nice to keep it to `None` but add a `status` field to give some hints.


>     * currently calling `threadpool_limits(1, 'tbb')` will silently do nothing if tbb4py is not installed. I think it's not nice. I would raise a warning. Do you agree ?

I agree that a warning should be issued if `threadpool_limits(1)` is called on a program where `libtbb` is linked without `tbb4py` and an `RuntimeError` if `threadpool_limits(1, 'tbb')` is called.

threadpoolctl.py
jeremiedbb: Actually we need the version of libtbb DSO. So I guess TBB_runtime_interface_version() is what we want. Is that right ?
Thanks

anton-malakhov: Yes, I believe so.

jeremiedbb:
@anton-malakhov is it intended that when using tbb as mkl threading layer, MKL_NUM_THREADS or mkl_set_num_threads have no effect ?

anton-malakhov:
@jeremiedbb This is up to MKL implementation of this layer. I guess it happens because it is harder to control the number of TBB threads used in a library. There is a choice - use a separate task_arena for all the library calls, which can control concurrency but adds overheads, or let application to control context in which TBB is executing. I guess they chose the later.

jeremiedbb:
Thanks

jeremiedbb:
Closing since it seems that TBB is not meant to have its threads controlled as easily as for the other libs. The threadpoolctl API expects things we can't guarantee with TBB.

anton-malakhov:
What's wrong with tbb::global_control knob to do this? https://spec.oneapi.io/versions/0.5.0/oneTBB/task_scheduler/tbb_global_control.html

jeremiedbb:
The issue I encountered is that I can lower the number of threads that TBB can use but then I can't restore it to its original value, which is expected since threadpoolctl is mainly meant to be used as a context manager

anton-malakhov:
When you create the global_control class instance, do you release it before/after you are creating new one? If  yes, then it might be a bug which worth filling

jeremiedbb:
Not sure, it's been a long time since I haven't touched this PR :)
Maybe I did something wrong, I can give it another try in a new PR. The code base has changed so much since then that it can't be synced with master anyway now.

PR #48, opened 2019-10-24
[MRG] ENH: expose threading layer of mkl

jeremiedbb:
side quest toward #47
it will allow to have a special behavior when using tbb as threading layer for mkl in our tests (e.g. with tbb, mkl_set_num_threads has no effect)


jeremiedbb:
I updated the change log. merging.

Issue #40, opened 2019-09-25
Deadlock when calling threadpool_limits from python threadpool


jeremiedbb:
In very specific conditions.

To reproduce: 
- environment with numpy, mkl
- clang compiler

`CC_OUTER_LOOP=clang continuous_integration/build_test_ext.sh`
(inner loop compiler is irrelevant)

```python
from threadpoolctl import threadpool_info 
import numpy as np 
from tests._openmp_test_helper.nested_prange_blas import check_nested_prange_blas 
from multiprocessing.pool import ThreadPool 
 
A = np.ones((1000, 10)) 
B = np.ones((100, 10)) 
 
# blas call within a prange
# remove this and no more deadlock
# Lower number of threads (4) and no more deadlock
check_nested_prange_blas(A, B, 4)

# Remove this and no more deadlock.
# Put it before check_nested_prange_blas and no more deadlock
np.dot(A, B.T) 
 
def func(i): 
    threadpool_info()
    # threadpool_info() can be replaced by
    # check_nested_prange_blas(A, B, 1)
    # the 1 is important
     
def test(): 
    pool = ThreadPool(2) 
    pool.map(func, range(2)) 
    pool.close() 
    pool.join()  # no join = no more deadlock
 
test() 
test()  # remove the second one and no more deadlock
```

@ogrisel @tomMoral

jeremiedbb:
Additional informations

- **libiomp** raises the following warning during the call to `check_nested_prange_blas`:
```
OMP: Warning #190: Forking a process while a parallel region is active is potentially unsafe.
```

- It's the call to `get_num_threads` of **libomp** in threadpool_info which causes the deadlock

ogrisel:
> libiomp raises the following warning during the call to check_nested_prange_blas:
> OMP: Warning #190: Forking a process while a parallel region is active is potentially unsafe.

This is unexpected no? This code does not seem to do any fork.

jeremiedbb:
`check_nested_parange_blas` is essentially:
```
for i in prange(n):
    blas_gemm(...)
```

If I interpret correctly the warning, the "activve parallel region" is the outer OpenMP (libomp) parallel region. And the "fork" is libiomp creating new threads (OpenMP thread creation is through fork, see https://en.wikipedia.org/wiki/OpenMP)

ogrisel:
[...]

PR #49, opened 2019-10-24
[MRG] Warn when llvm + intel loaded at the same time

jeremiedbb:
Following the discussion in #40 and in particular [https://github.com/joblib/threadpoolctl/issues/40#issuecomment-545377697](https://github.com/joblib/threadpoolctl/issues/40#issuecomment-545377697)

both implementations of OpenMP are actually the same and are incompatible in the sense that they do not accept having multiple copies at the same time. See this error message from llvm
``OMP: Hint This means that multiple copies of the OpenMP runtime have been linked into the program.``

(un)fortunately crashes should not occur very often outside of threadpoolctl since we are only able to create them by accessing the openmp api (get/set num threads, ...), not by mixing pranges and blas calls.

This PR adds the emission of a warning when trying to access threadpoolctl functionalities.

TODO:
 - [x] Add a page documenting all our discoveries about that and possible workarounds, and point the warning to that page

 - [x] test on macOS, especially the switch of threading layer for mkl

 - [x] add a test for the warning

 - [x] warn also in threadpool_limits

 - [x] What to do with all our clang ci jobs ?

### DIFF
--- a/.azure_pipeline.yml
+++ b/.azure_pipeline.yml
@@ -77,8 +77,7 @@ stages:
           VERSION_PYTHON: '3.7'
           CC_OUTER_LOOP: 'gcc'
           CC_INNER_LOOP: 'clang-8'
-        # Linux environment with numpy from conda-forge channel and
-        # heterogeneous OpenMP runtimes.
+        # Linux environment with numpy from conda-forge channel
         pylatest_conda_forge:
           PACKAGER: 'conda-forge'
           VERSION_PYTHON: '*'
--- a/.azure_pipeline.yml
+++ b/.azure_pipeline.yml
@@ -53,16 +53,23 @@ stages:
           BLAS: 'openblas'
           CC_OUTER_LOOP: 'clang-8'
           CC_INNER_LOOP: 'clang-8'
-        # Linux environment with MKL and Clang, known to be unsafe, to test the
-        # warning only.
+        # Linux environment with MKL and Clang (known to be unsafe for
+        # threadpoolctl) to only test the warning from multiple OpenMP.
         pylatest_conda_mkl_clang_gcc:
           PACKAGER: 'conda'
           VERSION_PYTHON: '*'
           BLAS: 'mkl'
           CC_OUTER_LOOP: 'clang-8'
           CC_INNER_LOOP: 'gcc'
-          MKL_THREADING_LAYER: 'INTEL'
           TESTS: 'libomp_libiomp_warning'
+        # Linux environment with MKL, safe for threadpoolctl.
+        pylatest_conda_mkl_gcc_gcc:
+          PACKAGER: 'conda'
+          VERSION_PYTHON: '*'
+          BLAS: 'mkl'
+          CC_OUTER_LOOP: 'gcc'
+          CC_INNER_LOOP: 'gcc'
+          MKL_THREADING_LAYER: 'INTEL'
         # Linux + Python 3.7 with numpy / scipy installed with pip from PyPI
         # and heterogeneous OpenMP runtimes.
         py37_pip_openblas_gcc_clang:
--- a/.azure_pipeline.yml
+++ b/.azure_pipeline.yml
@@ -53,14 +53,16 @@ stages:
           BLAS: 'openblas'
           CC_OUTER_LOOP: 'clang-8'
           CC_INNER_LOOP: 'clang-8'
-        # Linux environment to test the latest available dependencies and MKL.
-        pylatest_conda_mkl_gcc_gcc:
+        # Linux environment with MKL and Clang, known to be unsafe, to test the
+        # warning only.
+        pylatest_conda_mkl_clang_gcc:
           PACKAGER: 'conda'
           VERSION_PYTHON: '*'
           BLAS: 'mkl'
-          CC_OUTER_LOOP: 'gcc'
+          CC_OUTER_LOOP: 'clang-8'
           CC_INNER_LOOP: 'gcc'
           MKL_THREADING_LAYER: 'INTEL'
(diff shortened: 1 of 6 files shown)


ogrisel:
> What to do with all our clang ci jobs ?

Make sure we use numpy with openblas when testing under Linux with one of the clang compiled extension.

jeremiedbb:
> test on macOS, especially the switch of threading layer for mkl

there's no mkl init in numpy on macos. So mkl keeps intel threading layer even if llvm openmp is already loaded. This indicates that both libs are not incompatible as they are on linux. Moreover I haven't been able to reproduce any of the issue we encountered. It may be due (wild guess) to a difference in the dynamic loading process, but I really have no idea.

So only raise the warning on linux seems enough for now.

ogrisel:
This should be rebased on top of current master to use the new internal object oriented API.

jeremiedbb:
I rebased on top of master. Finally I did not add a test to check the warning since with this PR we'll avoid the mix libomp, libiomp on the ci. Codecov won't be happy once again :)

I added a page to document our OpenMP investigations. It's a first attempt, comments welcomed. Or even, feel free to edit directly.

multiple_openmp.md
```diff
@@ -0,0 +1,35 @@
...
+(Intel's math library) and a package using multi-threading with OpenMP
+(Scikit-learn, LightGBM, XGBoost, ...).
+
+From our experience, most OpenMP libraries can seamlessly coexist in a same
+program. For instance, on Linux, we never observed any issue between libgomp
+and libiomp, which is the most common mix (numpy with MKL + a package compiled
+with GCC, the most widely used C compiler).
+
+The only unrecoverable incompatibility we encountered is between libomp and
+libiomp, on Linux, manifested by crashes or deadlocks. It can happen even with
```
ogrisel: Maybe repeat the name of the matching compiler each time ... "between libomp (with clang/llvm) and libiomp (with ICC), on Linux, ...."

multiple_openmp.md
```diff
@@ -0,0 +1,35 @@
...
+the simplest OpenMP calls like getting the maximum number of threads that will
+be used in a subsequent parallel region. An explanation is that libomp is
```
ogrisel: We are not 100% sure this is the actual cause:

```suggestion
be used in a subsequent parallel region. A possible explanation is that libomp is
```

threadpoolctl.py
```diff
@@ -523,6 +525,24 @@ def _check_prefix(self, library_basename, filename_prefixes):
...
+    def _warn_if_incompatible_openmp(self):
+        """Raise a warning if llvm-OpenMP and intel-OpenMP are both loaded"""
+        if sys.platform != 'linux':
+            # Only raise the warning on linux
+            return
+
+        prefixes = [module.prefix for module in self.modules]
+        msg = textwrap.dedent(
+            """
+            Found intel OpenMP ('libiomp') and llvm OpenMP ('libomp') loaded at
```
```suggestion
            Found Intel OpenMP ('libiomp') and LLVM OpenMP ('libomp') loaded at
```

threadpoolctl.py
```diff
@@ -523,6 +525,24 @@ def _check_prefix(self, library_basename, filename_prefixes):
...
+            the same time. Both libraries are known to be incompatible.
```
ogrisel: and this can cause random crashes or deadlocks on Linux when loaded in the same Python program.

multiple_openmp.md
```diff
@@ -0,0 +1,35 @@
...
+actually a fork of libiomp causing name colliding for instance. Using
+threadpoolctl may crash your program in such a setting.
+
+Surprisingly, we never encountered this kind of issue on macOS, where this mix
+is the most frequent (Clang being the default C compiler on macOS).
```
ogrisel: We should have a section on possible workarounds:

- Building your OpenMP-enabled extensions with GCC (or ICC) instead of clang/LLVM under Linux.
- How to install numpy with openblas using conda from default channel or conda-forge.

.azure_pipeline.yml
```diff
@@ -53,30 +53,38 @@ stages:
...
-        # and heterogeneous openmp runtimes.
+        # and heterogeneous OpenMP runtimes.
         py37_pip_openblas_gcc_clang:
           PACKAGER: 'pip'
           VERSION_PYTHON: '3.7'
           CC_OUTER_LOOP: 'gcc'
           CC_INNER_LOOP: 'clang-8'
-        # Linux environment with numpy from conda-forge channel
+        # Linux environment with numpy from conda-forge channel and
+        # heterogeneous OpenMP runtimes.
```
ogrisel:
```suggestion
        # homogeneous OpenMP runtimes.
```

gcc inner & outer loops => homogeneous

tests/test_threadpoolctl.py
```diff
@@ -361,3 +362,29 @@ def test_blis_threading_layer():
...
+
+
+@pytest.mark.skipif(not cython_extensions_compiled,
+                    reason='Requires cython extensions to be compiled')
+def test_libomp_libiomp_warning(recwarn):
+    # Trigger the import of a potentially clang-compiled extension:
+    from ._openmp_test_helper import check_nested_openmp_loops  # noqa
+
+    # Trigger the import of numpy to potentially import Intel OpenMP via MKL
+    import numpy.linalg  # noqa
```
jeremiedbb: This is not necessary since we already import numpy for the whole module

ogrisel: explicit is better than implicit :)

ogrisel:
Thank you very much @jeremiedbb. I think it's time for a release. WDYT?

PR #51, opened 2019-11-06
[MRG] CI: Pin Python version when installing with pip

jeremiedbb:
When installing packages with pip we can't install the latest Python version because wheels may not yet be available. It's the case right now for scipy and Python 3.8.


jeremiedbb:
the coverage drop seems to come from the fact that numpy used to ship an old openblas which did not expose it's version. There's been a bug fix release a few days ago and now it's openblas 3.7. It means that now we don't have any job with an old openblas. I don't think we care.

PR #52, opened 2019-11-06
[MRG] CI: Remove unused legacy utilities from the test suite

jeremiedbb:
Removes 2 pytest fixtures from when threadpoolctl was developed in loky. We never use it. And if I remember correctly they were introduced mainly for debugging. They are not relevant anymore.

@tomMoral do you disagree ?


jeremiedbb:
CI is broken right now, see #51

PR #55, opened 2019-11-14
MNT Big refactoring

jeremiedbb:
(EDITED)
(I know it's a big change but I really think it will help maintenance and addition of new features.)

This PR proposes a huge refactoring of the whole module. My main motivation is that I find the code already quite hard to read, sometimes pretty messy, and adding new features, especially api specific ones, makes the code less and less readable. I find it more clear that all module related infos are grouped by internal api.

**Proposition**
- the public api does not change.  `threadpool_info` still return a list of dicts.
- internally we have new classes. 
  - `_ThreadpoolInfo` which contains (it's not a container :)) a list of the found modules.
  - Each module is a subclass of the `_Module` class, specific to each internal api (`_MKLModule`, `_OpenBLASModule`, `_OpenMPModule`, ....). Api specific features like `get_version`, `get_num_threads`, ... are grouped in those classes.

**Advantages**
Makes it easier and more readable to add api specific features, e.g. TBB support, MKL threading layer (already done), BLIS multithreading infos, ...
It also makes the tests simpler to read, because currently I find it's a nightmare. And also easier to adapt when adding new features.

**Backward compatibility**
The public interface is not changed.

**New feature (private)**
I just added a small feature. `_ThreadpoolInfo` has a `get_modules` method which allow to filter modules matching a pair `key: val`. I was tired of repeating the same dict comprehension stuff :)
Most of the changes in the test suite are due to that.

**Benchmarks**
`python benchmarks/bench_context_manager_overhead.py --import numpy --n-calls 10000`
    - master: 0.252 +/-0.063 ms (min of 5 trials)
    - this pr: 0.255 +/-0.075 ms  (min of 5 trials)
no slow down !

**Remarks**
The actual code is not longer (not really shorter either). There's more docstrings and comments :)


GaelVaroquaux:
As a general rule, I am afraid of subclasses of container objects: they
look like standard containers at first glance, but are not. It makes a
codebase hard to grasp for an outsider.

Would it be possible to use standard containers?

GaelVaroquaux:
> Would it be possible to use standard containers?

Or standard classes, maybe, to replace the subclass of dict?

jeremiedbb:
> Or standard classes, maybe, to replace the subclass of dict?

It can be a standard class with a dict in its attributes. But then it's not backward compatible in the sense that `threadpool_info` would then return a list of objects and you'd have to call object.some_attribute to get the module infos.

jeremiedbb:
> As a general rule, I am afraid of subclasses of container objects: they
look like standard containers at first glance, but are not. It makes a
codebase hard to grasp for an outsider.

I don't see how it's different from subclassing another class. Can you be more specific ?

GaelVaroquaux:
> I don't see how it's different from subclassing another class. Can you be more specific ?

It looks like a dictionary

jeremiedbb:
Do you have an example of how it's behavior could differ from a dict ?
(Note that I inherit from UserList and UserDict which are meant do be subclassed and behave 100% like list and dict.)

GaelVaroquaux:
> Do you have an example of how it's behavior could differ from a dict ?

First it will be a subclass of dict (and reply to isinstance dict), second it will come with many of the methods of dict (__getitem__, __len__, values) which will make it look like a dict. These are problems if it is not a dict.

jeremiedbb:
Yes but it's the point :) It is a dict, and I want it to be a dict !

GaelVaroquaux:
I had a quick look at the code. I am not completely sure why you want the
object to behave like a dictionary, and why a class is not enough? You'd
have to replace things like "module['set_num_threads']" by
module.set_num_thread, but I think that this is fine, no?

jeremiedbb:
I agree it's not mandatory and a standard class could do the job, but:
- it's not backward compatible: threadpool_info() would not return a list of dict anymore.
- print(threadpool_info()) would not print the infos, unless we implement __repr__/__str__ but in that case pprint wouldn't work anymore.

It's also fine. I personally like the idea that from the user's perspective it's a dict (with extra features internally). He can do whatever dict operation he wants. Implementing a custom class would require document it's attributes, it's methods, to tell the user what he can do with the result of threadpool_info().

ogrisel:
We could release threadpoolctl 2.0 if we want to break backward compat. I think it's fine to break backward compat now as we do not have that many users at this point.

jeremiedbb:
So besides this particular point (although I'm still confused about your mistrust towards container subclassing :D) what's your general opinion about moving to a more object oriented implementation ? Do you find my concerns valid points ? Do you think it's overkill ?

GaelVaroquaux:
> what's your general opinion about moving to a more object oriented implementation ?

I can see the value.

> Do you think it's overkill ?

It's not overkill if it simplifies the code base and makes it easier to
support more cases.

jeremiedbb:
@GaelVaroquaux I thought about that question of subclassing again. I tried different implementations and all feel wrong. I think the appropriate object to inherit from should be deduced from our desired behavior of `threadpool_info()`.

- We want `threadpool_info()` to directly return all info such that we can do `print(threadpool_info())`
 
- We want it to be an iterable so that we can do `for module in threadpool_info()` so the natural way to represent it is a list.

- For each module the info is the value of a few properties of the underlying shared library, so the natural way to represent it is a dict.

All this can be achieved with an attribute `modules` which stores the list of modules and an attribute `info` which stores the dict of info but if you have a class with an attribute of some <type> and you only want to access that class through it's attribute then your class should instead inherit from that <type>.

ogrisel:
I am also not a fan of subclassing base container types because I find it surprising.

That being said it's important to have an informative `__repr__` / `__str__` for the output of  threadpool_info.

I think what we need are dataclasses but they are not available before we drop support for Python 3.6 unfortunately. So we are stuck with traditional classes.

Also `_ThreadpoolInfo` is marked as private but since instances are returned by the main public function, it's de-facto part of the public API. So we should name it as such (not leading `_`).

threadpoolctl.py
```diff
@@ -16,16 +16,11 @@
 import ctypes
 import warnings
 from ctypes.util import find_library
+from abc import ABC, abstractmethod
 
-__version__ = '1.2.0.dev0'
+__version__ = "1.2.0.dev0"
```
ogrisel: Because this is a big refactoring we might want to target `__version__ = "2.0.0.dev0"` once this PR is merged, even if the backward compatibility should be preserved.

threadpoolctl.py
```diff
@@ -539,29 +137,36 @@ class threadpool_limits:
...
-    The `user_api` parameter selects particular APIs of libraries to limit.
-    Used only if `limits` is an int. If it is None, this function will apply to
-    all supported libraries. If it is "blas", it will limit only BLAS supported
-    libraries and if it is "openmp", only OpenMP supported libraries will be
-    limited. Note that the latter can affect the number of threads used by the
-    BLAS libraries if they rely on OpenMP.
+    Parameters
+    ----------
+    limits : int, dict or None (default=None)
+        The maximal number of thread that can be used in thread pools
```
```suggestion
        The maximal number of threads that can be used in thread pools
```

threadpoolctl.py
```diff
@@ -539,29 +137,36 @@ class threadpool_limits:
...
+
+        - If int, sets the maximum number of thread to `limits` for each
```
ogrisel: I think we should use ["number of" + plural](https://ell.stackexchange.com/questions/87072/is-a-number-of-plural-or-singular). I know that this mistake was already present in the previous state of the code base, but we might as well fix it while we are at it.

```suggestion
        - If int, sets the maximum number of threads to `limits` for each
```

threadpoolctl.py
```diff
@@ -539,29 +137,36 @@ class threadpool_limits:
...
+          library selected by `user_api`.
+
+        - If it is a dictionary `{{key: max_threads}}`, this function sets a
+          custom maximum number of thread for each `key` which can be either a
```
```suggestion
          custom maximum number of threads for each `key` which can be either a
```

jeremiedbb:
So I dropped the container subclassing. `_ThreadpoolInfo` now has a `modules` attribute which is a list of `_Modules`. It also has a `todicts` method which returns the threadpool info represented as a list of dicts (for backward compat, avoid implementing repr and str ourselves, natively supported by pprint).

One argument in your favor, the benchmarks now show no slow down at all \o/. It's because in the first implem I inherited from `UserList` and `UserDict` which are meant to be subclassed but are not as efficient as builtin list and dict.

threadpoolctl.py
```diff
@@ -599,6 +211,520 @@ def get_original_num_threads(self):
...
+        """Load a windows DLL"""
+        dll = cls._system_libraries.get(dll_name)
+        if dll is None:
+            dll = ctypes.WinDLL("{}.dll".format(dll_name))
+            cls._system_libraries[dll_name] = dll
+        return dll
+
+    @classmethod
+    def _realpath(cls, filepath, cache_limit=10000):
+        """Small caching wrapper around os.path.realpath to limit system calls
```
ogrisel: Maybe make it explicit that we use a cls level cache instead of an instance level cache because we can safely assume that the filepath of loaded charged library will never change during the lifetime of a program.

threadpoolctl.py
```diff
@@ -599,6 +211,520 @@ def get_original_num_threads(self):
...
+
+        Return None if none matches.
+        """
+        for prefix in filename_prefixes:
+            if library_basename.startswith(prefix):
+                return prefix
+        return None
+
+    @classmethod
+    def _get_libc(cls):
```
ogrisel: Maybe make it explicit that we use a class level cache instead of an instance level cache because shared libraries with a fixed name cannot be unloaded and reloaded during the lifetime of a program.

threadpoolctl.py
```diff
@@ -599,6 +211,520 @@ def get_original_num_threads(self):
...
+            libc_name = find_library("c")
+            if libc_name is None:  # pragma: no cover
+                return None
+            libc = ctypes.CDLL(libc_name, mode=_RTLD_NOLOAD)
+            cls._system_libraries["libc"] = libc
+        return libc
+
+    @classmethod
+    def _get_windll(cls, dll_name):
+        """Load a windows DLL"""
```
ogrisel: Maybe make it explicit that we use a class level cache instead of an instance level cache because shared libraries with a fixed name cannot be unloaded and reloaded during the lifetime of a program.

threadpoolctl.py
```diff
@@ -599,6 +211,520 @@ def get_original_num_threads(self):
...
+
+
+@_format_docstring(
+    PREFIXES=", ".join('"{}"'.format(prefix) for prefix in _ALL_PREFIXES),
+    USER_APIS=", ".join('"{}"'.format(api) for api in _ALL_USER_APIS),
+    BLAS_LIBS=", ".join(_ALL_BLAS_LIRARIES),
+    OPENMP_LIBS=", ".join(_ALL_OPENMP_LIBRARIES))
+class _ThreadpoolInfo():
+    """Collection of all supported modules that have been found
+
```
ogrisel: Maybe add a not to explain that the object oriented API of `_ThreadpoolInfo` and its model is private and the public API (i.e. the "threadpool_info" function) only exposes the "list of dicts" representation returned by the `.todicts` method.

threadpoolctl.py
```diff
@@ -599,6 +211,520 @@ def get_original_num_threads(self):
...
+        """Return the maximum number of threads available to use"""
+        pass  # pragma: no cover
+
+    @abstractmethod
+    def set_num_threads(self, num_threads):
+        """Set the maximum number of threads to use"""
+        pass  # pragma: no cover
+
+    @abstractmethod
+    def _get_extra_info(self):
```
ogrisel: Why is this abstract method private if all others are "public". The whole class is object oriented module API is private anyway so I think it would be fine to remove the leading `_`? WDYT?

threadpoolctl.py
```diff
@@ -40,19 +35,17 @@
...
 
 
 class _dl_phdr_info(ctypes.Structure):
     _fields_ = [
         ("dlpi_addr",  _SYSTEM_UINT),       # Base address of object
-        ("dlpi_name",  ctypes.c_char_p),   # path to the library
-        ("dlpi_phdr",  ctypes.c_void_p),   # pointer on dlpi_headers
+        ("dlpi_name",  ctypes.c_char_p),    # path to the library
+        ("dlpi_phdr",  ctypes.c_void_p),    # pointer on dlpi_headers
         ("dlpi_phnum",  _SYSTEM_UINT_HALF)  # number of element in dlpi_phdr
```
```suggestion
        ("dlpi_phnum",  _SYSTEM_UINT_HALF)  # number of elements in dlpi_phdr
```

threadpoolctl.py
```diff
@@ -599,6 +211,520 @@ def get_original_num_threads(self):
...
+        return self.todict() == other.todict()
+
+    def todict(self):
+        """Return relevant info wrapped in a dict"""
+        return {"user_api": self.user_api,
+                "internal_api": self.internal_api,
+                "prefix": self.prefix,
+                "filepath": self.filepath,
+                "version": self.version,
+                "num_threads": self.num_threads}
```
ogrisel: Since we are now testing a lot based on the private API, I think we should have dedicated, non-regression, unit tests for the `_Module.todict` and `_ThreadPoolInfo.todicts` methods to avoid introducing regressions silently.

ogrisel:
Nice! the new "No test always skipped" CI job has detected a meta-coverage regression:

https://dev.azure.com/joblib/threadpoolctl/_build/results?buildId=170

jeremiedbb:
> Nice! the new "No test always skipped" CI job has detected a meta-coverage regression:

Yes I'm looking at it. It's weird since I didn't touch to the azure config.
One weird thing is that we have openblas instead of mkl in a mkl job (although unrelated to the fail because it was already there in last run on master, just checked). We might need to put back what I removed in #52 and actually use it :)

jeremiedbb:
And there's a bug in the meta-test :( because `test_original_num_threads` is not skipped in every job (manually checked)

ogrisel:
Or maybe it's a bug in the azure workspace share folder that was not completely up to date when the script was executed? Maybe you could make the meta-test script more verbose.

jeremiedbb:
real bug from my side :/
I'm making it more verbose. PR soon(tm).

ogrisel:
This can be updated to re-run the tests with the current master since I just merged #58.

jeremiedbb:
yep. It should fix one of the always skipped test but it requires #59 to fix the other one
What happen is I fixed `test_get_original_num_threads` in this PR which checks that a warning is raised. Pytest records that in the results and it ended up as a false positive.

I'll merge #59 in this PR after this build to check that is effectively solves the issue

jeremiedbb:
it's working :)

ogrisel:
It's weird, codecov seems to report that the windows tests were not executed:

https://codecov.io/gh/joblib/threadpoolctl/compare/1e27bdf92d5934e3544f6c9a4a62a3f808dfcec0...4e2c2cd5395f1855e245e755f8d03644a588c34a/changes

jeremiedbb:
Yes I saw but they have been... I never trusted codecov anyway :)
Moreover I checked that the upload to codecov step was successful as well...

threadpoolctl.py
jeremiedbb: My policy for _method in a private class is a method intended to be used exclusively inside the class (a private method).

jeremiedbb:
I pushed an empty commit to trigger the ci and the coverage worked fine. It was just yet another codecov surprise

tests/test_threadpoolctl.py
```diff
@@ -213,16 +231,17 @@ def test_openmp_nesting(nthreads_outer):
 
 
 def test_shipped_openblas():
-    all_openblases = [ctypes.CDLL(path) for path in libopenblas_paths]
-    original_num_threads = [blas.openblas_get_num_threads()
-                            for blas in all_openblases]
+    # checks that OpenBLAS effectively uses the number of threads requested by
+    # the context manager
+    original_info = _threadpool_info()
```
tomMoral:
```suggestion
    original_infos = _threadpool_info()
```
To be consistent with the previous one, it makes it clear it is the same thing.
Or maybe change it before to have the same name as in `ctx._original_info`.

threadpoolctl.py
```diff
@@ -599,6 +211,529 @@ def get_original_num_threads(self):
...
+
+    def _set_threadpool_limits(self):
+        """Change the maximal number of threads in selected thread pools.
+
+        Return a list with all the supported modules that have been found
+        matching `self._prefixes` and `self._user_api`.
+        """
+        if self._limits is None:
+            return None
+        else:
```
tomMoral: nit: I usually prefer to not have the `else` in this case as the body of the function is included in it.

threadpoolctl.py
tomMoral: Shouldn't the `todict` method also return the extra info? This could be done by storing the `extra_keys` and looping on them here.

threadpoolctl.py
```diff
@@ -599,6 +211,529 @@ def get_original_num_threads(self):
...
+    @abstractmethod
+    def _get_extra_info(self):
+        """Add additional module specific information"""
+        pass  # pragma: no cover
+
+
+class _OpenBLASModule(_Module):
+    """Module class for OpenBLAS"""
+    user_api = "blas"
+    internal_api = "openblas"
```
tomMoral: This feels redundant with the `_SUPPORTED_MODULES` list of dictionary. Maybe it would be easier to simply have a list of classes which contains the info. This way, the info wouldn't be duplicated.

tests/test_threadpoolctl.py
```diff
@@ -24,105 +22,124 @@ def effective_num_threads(nthreads, max_threads):
...
                 continue
-            assert module["num_threads"] == 1
+            assert module.num_threads == 1
     finally:
         # Restore the original limits so that this test does not have any
         # side-effect.
         threadpool_limits(limits=original_infos)
 
-    assert threadpool_info() == original_infos
+    assert _threadpool_info() == original_infos
```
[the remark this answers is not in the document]
ogrisel: It depends on the test. For tests that can be simplified significantly by using the private API I think it's ok to use it. Maybe we could add a comment to explain why we test against the private API in those cases.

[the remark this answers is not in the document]
tomMoral: Yes I agree for tests that are simplified a lot with the use of `_threadpool_info`. I only meant for tests that have been modified only to accomodate with the use of `is_old_openblas`.

tests/test_threadpoolctl.py
jeremiedbb: I went for info everywhere, see [https://blog.harwardcommunications.com/2010/11/09/information/](https://blog.harwardcommunications.com/2010/11/09/information/) :)

jeremiedbb:
> Shouldn't the todict method also return the extra info? This could be done by storing the extra_keys and looping on them here.

Yes, good catch. It was actually a bug. I fixed it and improved the tests to actually check it

threadpoolctl.py
jeremiedbb: I reworked that a little bit. This info is now only present in _SUPPORTED_MODULES and passed to the constructor of the module.

tests/test_threadpoolctl.py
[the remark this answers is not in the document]
jeremiedbb: I agree that some tests are not simplified with the new object api. I chose to use the private api in all existing tests for consistency because otherwise I think it might be confusing to have some tests using the public and some tests using the private api with no obvious reason to choose one over the other. Of course, in addition, I made new tests to check consistency between the public and private api.

tests/test_threadpoolctl.py
tomMoral: Perfect!

PR #57, opened 2019-11-19
MNT CI Add a meta-test ensuring no test is always skipped 

jeremiedbb:
I had the feeling that `test_multiple_shipped_openblas` was skipped in every ci job.
Good news it's not :)

However I think it's still useful to make sure all our tests are run in at least 1 job, and looping manually on azure is quite cumbersome.

The test is a python script ran after all regular jobs have succeeded, looping through all pytest results and failing is one or more test has been skipped in every job. 
I added the script under continuous_integration and not under test because it only make sens in a ci setup.


tests/utils.py
```diff
@@ -1,27 +1,18 @@
 import os
-import pytest
 from glob import glob
 
 
-def skip_func(msg):
-    def test_func(*args, **kwargs):
-        pytest.skip(msg)
-    return test_func
-
```
jeremiedbb: I had to change that because of it ignores parametrized tests, which means in some job you'd get
```
tests/test_threadpoolctl.py::test_openmp_nesting[None] PASSED
tests/test_threadpoolctl.py::test_openmp_nesting[1] PASSED
tests/test_threadpoolctl.py::test_openmp_nesting[2] PASSED
tests/test_threadpoolctl.py::test_openmp_nesting[4] PASSED
```
and in other jobs you'd get
```
tests/test_threadpoolctl.py::test_openmp_nesting SKIPPED
```
which makes merging all test files painful.

Using skipif makes pytest skip the tests individually.

ogrisel:
LGTM, I just pushed a commit to test the test infra. I will revert and merge if everything goes well.

ogrisel:
Nice! Let me clean-up and merge.

PR #58, opened 2019-11-22
[MRG] MNT use new syntax to select blas with conda

jeremiedbb:
conda install numpy is less predictable than it used to be w.r.t blas.
in python < 3.8 it will install mkl; in python 3.8 it's openblas (for now ?)

We can be sure of which implem of blas we will get with the following syntax
`conda install numpy blas[build=mkl]`

An alternative syntax is `blas=*=mkl` but I prefer the first one.

@ogrisel Fixes the first part of the issue in #55

### DIFF
--- a/continuous_integration/install.cmd
+++ b/continuous_integration/install.cmd
@@ -18,7 +18,7 @@ python --version
 pip --version
 
 @rem Install dependencies with either conda or pip.
-if "%PACKAGER%" == "conda" (%CONDA_INSTALL% numpy=1.15 scipy pytest cython blas[build=%BLAS%])
+if "%PACKAGER%" == "conda" (%CONDA_INSTALL% numpy scipy pytest cython)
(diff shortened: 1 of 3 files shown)


jeremiedbb:
I took the opportunity to refresh a bit our jobs:
- change py37 on windows to pylatest so that we can test python 3.8
- change one of the pylatest on mac to py35 so that we test our min python on mac
This way we have more coverage of the python versions on each platform.

continuous_integration/install.cmd
```diff
@@ -18,7 +18,7 @@ python --version
 pip --version
 
 @rem Install dependencies with either conda or pip.
-if "%PACKAGER%" == "conda" (%CONDA_INSTALL% numpy=1.15 scipy pytest cython)
+if "%PACKAGER%" == "conda" (%CONDA_INSTALL% numpy scipy pytest cython)
```
jeremiedbb: I also unpinned the numpy version here. It was probably added at some point for a good reason but there's no explanation and it works without it.

PR #59, opened 2019-11-22
[MRG] MNT Fix meta-test distinguish skipped from other records

jeremiedbb:
Actually check that the test is skipped.
Before that, a false positive could be due to a warning raised in a test for instance.

Also made the meta test way more verbose

### DIFF
--- a/continuous_integration/check_no_test_skipped.py
+++ b/continuous_integration/check_no_test_skipped.py
@@ -24,16 +24,13 @@ for name in os.listdir(base_dir):
         # All tests are identified by the xml tag testcase.
         for test in root.iter("testcase"):
             test_name = test.attrib["name"]
-            if test_name not in always_skipped:
-                always_skipped[test_name] = True
-
-            for child in test:
-                if child.tag == "skipped":
-                    print("    -", test_name)
-                    always_skipped[test_name] &= True
-                    break
+            skipped = any(child.tag == "skipped" for child in test)
+            if skipped:
+                print("    -", test_name)
+            if test_name in always_skipped:
+                always_skipped[test_name] &= skipped
             else:
-                always_skipped[test_name] = False
+                always_skipped[test_name] = skipped
 
 print("\n------------------------------------------------------------------\n")
 fail = False
--- a/continuous_integration/check_no_test_skipped.py
+++ b/continuous_integration/check_no_test_skipped.py
@@ -16,18 +16,25 @@ aggregated_results = {}
 for name in os.listdir(base_dir):
     # all test result files are in /base_dir/jobs.*/ dirs
     if name.startswith("stage1."):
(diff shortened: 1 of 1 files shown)


continuous_integration/check_no_test_skipped.py
```diff
@@ -16,18 +16,26 @@
...
+            print("  -", test_name)
+
+            for child in test:
+                print("    -", child.tag)
+                if child.tag == "skipped":
+                    aggregated_results[test_name] |= False
+                    break
             else:
-                aggregated_results[test_name] |= not bool(len(test))
+                aggregated_results[test_name] |= True
```
ogrisel:
```suggestion
                aggregated_results[test_name] = True
```

would be even simpler right?

jeremiedbb: yes :)

jeremiedbb:
> Printing all the tests is going to be really verbose. Wouldn't it be enough to only print the (sorted) skipped test for each jobs?

I guess we can do that now that I've understood what was going on

continuous_integration/check_no_test_skipped.py
```diff
@@ -11,26 +11,34 @@
...
+                    break
             else:
-                aggregated_results[test_name] |= not bool(len(test))
+                always_skipped[test_name] = False
 
+print("\n------------------------------------------------------------------\n")
 fail = False
-for test, result in aggregated_results.items():
-    if not result:
+for test, skipped in always_skipped.items():
```
ogrisel: Not sure if necessary, but never hurts:

```suggestion
for test, skipped in sorted(always_skipped.items()):
```

jeremiedbb: hum why do you want to reorder the tests ? I find it easier is they keep the order they are run in the test suite.

ogrisel: Alright.

continuous_integration/check_no_test_skipped.py
```diff
@@ -11,26 +11,34 @@
...
+                always_skipped[test_name] = True
+
+            for child in test:
+                if child.tag == "skipped":
+                    print("    -", test_name)
+                    always_skipped[test_name] &= True
+                    break
             else:
-                aggregated_results[test_name] |= not bool(len(test))
+                always_skipped[test_name] = False
```
ogrisel: What about?

```python
        for test in root.iter("testcase"):
            test_name = test.attrib["name"]
            skipped = any(child.tag == "skipped" for child in test)
            if test_name in always_skipped:
                always_skipped &= skipped
            else:
                always_skipped = skipped
```

ogrisel:
Merged! Thanks very much for the fix @jeremiedbb.

PR #60, opened 2019-11-29
[MRG] ENH add threading layer for BLIS and OpenBLAS

jeremiedbb:
Show information about how BLIS and OpenBLAS were built. Both can be built with multithreading disabled, multithreading with pthreads or multithreading with OpenMP.

I'm not sure about the name, because it's not exactly the same kind as mkl's threading layer. Maybe "multithreading" = {"disabled", "pthreads", "openmp"} would be better. What do you think ?

### DIFF
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -637,14 +637,14 @@ class _OpenBLASModule(_Module):
         return set_func(num_threads)
 
     def _get_extra_info(self):
-        self.threading_layer = self.get_threading_layer()
+        self.multithreading = self.get_multithreading()
 
     def get_threading_layer(self):
         """Return the threading layer of OpenBLAS"""
-        threading_layer = self._dynlib.openblas_get_parallel()
-        if threading_layer == 2:
+        multithreading = self._dynlib.openblas_get_parallel()
+        if multithreading == 2:
             return "openmp"
-        elif threading_layer == 1:
+        elif multithreading == 1:
             return "pthreads"
         return "disabled"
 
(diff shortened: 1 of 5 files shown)


tomMoral:
I think using `'multithreading'` would be nicer if it is not the same thing as MKL, it would avoid confusion.

jeremiedbb:
The lack of coverage is due to the fact that we don't have OpenBLAS or BLIS built with all possible multithreading options. But I don't think we want to add many jobs just to cover that. For instance we don't want to build openmp in our ci, and no packager ever ships numpy with openblas built with openmp.

jeremiedbb:
> I think using 'multithreading' would be nicer if it is not the same thing as MKL, it would avoid confusion.

I didn't need much to be convinced :)

ogrisel:
Personally I liked "threading_layer" better. Why would this be a problem? "threading layer" is a generic enough name and it is scoped in the context of a specific lib.

jeremiedbb:
After irl discussion I put back the threading layer name.

(@pierreglaser will blame me for my weak opinion weakly held :D)

threadpoolctl.py
```diff
@@ -637,7 +637,16 @@ def set_num_threads(self, num_threads):
...
 
     def _get_extra_info(self):
-        pass
+        self.threading_layer = self.get_threading_layer()
+
+    def get_threading_layer(self):
+        """Return the threading layer of OpenBLAS"""
+        threading_layer = self._dynlib.openblas_get_parallel()
+        if threading_layer == 2:
+            return "openmp"
```
ogrisel: This is not covered by any test but I suppose it would be too much of a hassle to setup a dedicated CI config just for this as AFAIK there is no binary package for OpenBLAS built with openmp.

PR #61, opened 2019-12-04
[MRG] MNT CI Turn Warnings into Errors

jeremiedbb:
- turn all warnings into errors in CI
- filter the warning from `get_original_num_threads`
- clean up a print in the tests which should not be there
- display the versions and threadpool_info before the tests on windows as it's already done on linux

### DIFF
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -302,8 +302,10 @@ def test_nested_prange_blas(nthreads_outer):
     assert original_info == _threadpool_info()
 
 
-# the method `get_original_num_threads` raises a UserWarning in one CI job. It
-# is expected so we can safely filter it.
+# the method `get_original_num_threads` raises a UserWarning due to different
+# num_threads from libraries with the same `user_api`. It will be raised only
+# in the CI job with 2 openblas (py37_pip_openblas_gcc_clang). It is expected
+# so we can safely filter it.
 @pytest.mark.filterwarnings("ignore::UserWarning")
 @pytest.mark.parametrize("limit", [1, None])
 def test_get_original_num_threads(limit):
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -302,6 +302,8 @@ def test_nested_prange_blas(nthreads_outer):
     assert original_info == _threadpool_info()
 
 
+# the method `get_original_num_threads` raises a UserWarning in one CI job. It
+# is expected so we can safely filter it.
 @pytest.mark.filterwarnings("ignore::UserWarning")
 @pytest.mark.parametrize("limit", [1, None])
 def test_get_original_num_threads(limit):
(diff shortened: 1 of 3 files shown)


tests/test_threadpoolctl.py
```diff
@@ -302,6 +302,7 @@ def test_nested_prange_blas(nthreads_outer):
     assert original_info == _threadpool_info()
 
 
+@pytest.mark.filterwarnings("ignore::UserWarning")
```
ogrisel: Could you please add a comment that explains both why we UserWarning to be raised in this test and why this is ok to ignore it here (insert of asserting that this warning is raised in 100% of the case for instance).

tests/test_threadpoolctl.py
```diff
@@ -313,7 +314,6 @@ def test_get_original_num_threads(limit):
         original_info = _threadpool_info()
         with threadpool_limits(limits=limit, user_api="blas") as threadpoolctx:
             original_num_threads = threadpoolctx.get_original_num_threads()
-            print(original_num_threads)
```
ogrisel: It would be great to find a way to get pytest to fail if any test prints anything to stdout / stderr. But don't worry if this is not easy to implement. It's just a nice to have.

jeremiedbb: I only found a way to check the stdout/stderr within each test. Not a reasonable solution...

tests/test_threadpoolctl.py
```diff
@@ -302,6 +302,8 @@ def test_nested_prange_blas(nthreads_outer):
     assert original_info == _threadpool_info()
 
 
+# the method `get_original_num_threads` raises a UserWarning in one CI job. It
+# is expected so we can safely filter it.
```
ogrisel: Could you please be more precise? Which warning? What CI configuration is causing it?

tests/test_threadpoolctl.py
```diff
@@ -302,6 +302,9 @@ def test_nested_prange_blas(nthreads_outer):
...
+@pytest.mark.filterwarnings("ignore::UserWarning")
```
jeremiedbb: Is there a way to filter a warning based on the expected warning message ?
I tried something like "ignore:Found" because 'Found' is in the warning message but it does not work.

ogrisel: You can use the context manager: 

https://docs.pytest.org/en/3.0.5/recwarn.html#recwarn

and then make assertions, ignore or raise based on the warning message contents.

jeremiedbb: yes but then pytest fails if there's no warning inside the context manager.

ogrisel: The `recwarn` fixture I linked above does not fail if so warning is raised.

ogrisel: But this is fine as it is anyway, merging.

PR #65, opened 2020-01-13
[MRG] Simple Command Line Interface

ogrisel:
I think it's helpful to expose a simple command-line interface for runtime introspection.

Please let me know what you think.


ogrisel:
I am working to make it more convenient when a package is missing.

ogrisel:
Alright, now we can put things like the following command in your CI script after installing stuff and before running pytest:

```
$ python -m threadpoolctl -i numpy scipy.linalg sklearn xgboost lightgbm
WARNING: could not import lightgbm
[
  {
    "filepath": "/home/ogrisel/miniconda3/envs/pylatest/lib/python3.7/site-packages/numpy/.libs/libopenblasp-r0-34a18dc3.3.7.so",
    "prefix": "libopenblas",
    "user_api": "blas",
    "internal_api": "openblas",
    "version": "0.3.7",
    "num_threads": 4,
    "threading_layer": "pthreads"
  },
  {
    "filepath": "/home/ogrisel/miniconda3/envs/pylatest/lib/libopenblasp-r0.3.7.so",
    "prefix": "libopenblas",
    "user_api": "blas",
    "internal_api": "openblas",
    "version": "0.3.7",
    "num_threads": 4,
    "threading_layer": "pthreads"
  },
  {
    "filepath": "/home/ogrisel/miniconda3/envs/pylatest/lib/libgomp.so.1.0.0",
    "prefix": "libgomp",
    "user_api": "openmp",
    "internal_api": "openmp",
    "version": null,
    "num_threads": 4
  }
]
```

As you see if some libraries are not installed on some CI jobs, you just get a warning in stderr instead of a failure.

tests/test_threadpoolctl.py
```diff
@@ -388,3 +390,45 @@ def test_libomp_libiomp_warning(recwarn):
...
+
+
+def test_command_line_empty():
+    output = subprocess.check_output(
+        "python -m threadpoolctl".split())
+    assert json.loads(output.decode("utf-8")) == []
+
+
+def test_command_line_command_flag():
+    pytest.importorskip("numpy")
```
tomMoral: Nice new helper from `pytest`!

tests/test_threadpoolctl.py
```diff
@@ -388,3 +390,45 @@ def test_libomp_libiomp_warning(recwarn):
...
+        "numpy",
+        "scipy.linalg",
+        "invalid_package",
+        "numpy.invalid_sumodule",
+    ], capture_output=True, check=True, encoding="utf-8")
+    cli_info = json.loads(result.stdout)
+
+    this_process_info = threadpool_info()
+    for module in cli_info:
+        assert module in this_process_info
```
tomMoral: I am unsure how the output will be the same here due to `scipy.linalg`. Is it imported in the `pytest` process? If not, it might have different modules no? Maybe it would be better to import it here to make sure we get the same result?

README.md
```diff
@@ -32,6 +32,37 @@ oversubscription issues.
...
+    "user_api": "openmp",
+    "internal_api": "openmp",
+    "version": null,
+    "num_threads": 4
+  }
+]
+```
+
+The JSON information is written on STDOUT. If some of the packages are missing,
+a warning message is displayed on STDERR.
```
tomMoral: Maybe also document the `-c` option?

README.md
```diff
@@ -32,6 +32,37 @@ oversubscription issues.
 
 ## Usage
 
+### Command Line Interface
```
jeremiedbb: I think the result should be the same as in the `runtime introspection` section otherwise it might be confusing to see different libraries when using the cli and when using the function.

README.md
ogrisel: It's already documented in the `-h` flag. I don't want the readme to be too verbose.

tests/test_threadpoolctl.py
ogrisel: I don't understand: how could the scipy imported by a subprocess call be different from the scipy imported from the parent process?

README.md
ogrisel: Alright.

README.md
tomMoral: ok

tests/test_threadpoolctl.py
tomMoral: Sorry my question was unclear. My question was: is `scipy.linalg` imported in the parent process. But upon further reading, it is at the beginning of this file if it is present. So nevermind!

ogrisel: And the tests pass :)

ogrisel:
Merged, thanks for the reviews.

PR #66, opened 2020-01-15
Use the CLI instead of continuous_integration/display_versions.py

ogrisel:
Let's eat our own dog food.


ogrisel:
> I guess the coverage decrease is just due to the removal of some line of codes?

It was caused by some tests that had not completed yet. Now it's fine.

But please do not merge, I would like to investigate other things in that CI.

ogrisel:
@jeremiedbb This is somewhat unrelated but my last commit improves the documentation on known limitations.

README.md
```diff
@@ -138,12 +138,27 @@ with threadpool_limits(limits=1, user_api='blas'):
...
+  libgomp from GCC and libomp from clang/llvm or libiomp from ICC).
+
+  See the `test_openmp_nesting` function in [tests/test_threadpoolctl.py](
+  https://github.com/joblib/threadpoolctl/blob/master/tests/test_threadpoolctl.py)
+  for an example. More information can be found at:
+  https://github.com/jeremiedbb/Nested_OpenMP
+
+  Note however that this problem does not happen when `threadpool_limits` is
+  used to limit the number of threads used internally by BLAS calls that are
+  them-selves nested under OpenMP parallel loops as long as `threadpool_limits`
```
jeremiedbb:
```suggestion
  themselves nested under OpenMP parallel loops as long as `threadpool_limits`
```

README.md
```diff
@@ -138,12 +138,27 @@ with threadpool_limits(limits=1, user_api='blas'):
...
+  is called prior to entering the outer parallel OpenMP region. This works as
```
jeremiedbb: I'm not sure this is true anymore, since the deadlock issues were due to the libiomp/libomp interactions on linux which we stopped supporting.

tests/test_threadpoolctl.py
```diff
@@ -373,7 +375,7 @@ def test_libomp_libiomp_warning(recwarn):
     from ._openmp_test_helper import check_nested_openmp_loops  # noqa
 
     # Trigger the import of numpy to potentially import Intel OpenMP via MKL
-    import numpy.linalg  # noqa
+    pytest.importorskip("numpy.linalg")
```
jeremiedbb: there's no way we can reach this part if numpy is not present since the test is skipped if scipy is not present.

ogrisel:
I do not understand why the coverage has decreased this time...

tests/test_threadpoolctl.py
```diff
@@ -263,6 +263,8 @@ def test_multiple_shipped_openblas():
 
 
 @pytest.mark.skipif(scipy is None, reason="requires scipy")
+@pytest.mark.skipif(not cython_extensions_compiled,
+                    reason='Requires cython extensions to be compiled')
```
jeremiedbb: Why is that necessary now ?

ogrisel: If you do not build the extension locally, no?

tests/test_threadpoolctl.py
ogrisel: There is no dependency on scipy for this one. `check_nested_openmp_loops` has no dependency on numpy/scipy.

README.md
ogrisel: I am not sure I understand.

The issue described at the beginning of this itemized section is still there. I get XFAILED for `test_openmp_nesting[2]` on my local Linux machine if I build the inner or outer loop with clang-8 and gcc for the other.

The OpenMP / BLAS case, tested via `check_nested_prange_blas`  works however. This is the case we have in scikit-learn. But then `threadpool_limit(1, user_api="blas")` is called prior to entering the outer parallel loop.

jeremiedbb: I did not point only the line I was referring to :/
My remark was reffering to `is called prior to entering the outer parallel`. Putting apart the linux/llvm/intel case, it should not be an issue to call threadpoolctl inside the outer loop.

tests/test_threadpoolctl.py
jeremiedbb: right, I mixed 2 tests...

README.md
ogrisel: Ok, I fixed this by removing this misleading/inaccurate information.

PR #74, opened 2020-06-02
[MRG] Rework test_openmp_nesting to be conda-forge friendly

jeremiedbb:


tests/utils.py
```diff
@@ -32,7 +38,40 @@
...
+    env = os.environ.copy()
+    try:
+        env["PYTHONPATH"] = os.pathsep.join([pythonpath, env["PYTHONPATH"]])
+    except KeyError:
+        env["PYTHONPATH"] = pythonpath
+
+    try:
+        with open(filename, "wb") as f:
+            f.write(src.encode("utf-8"))
+        cmd = [sys.executable, filename]
```
ogrisel, line 73: I think you could have used `cmd = [sys.executable, "-m", "threadpoolctl" + "-i" + module_to_import]`
and then read the result with `json.loads`.

jeremiedbb: much simpler indeed :)

PR #75, opened 2020-07-24
fix python executable in tests

MeggyCal:
As Python2 is dead, there can be systems which simply do not have "python" binary.


jeremiedbb (approved):

jeremiedbb:
Thanks @MeggyCal

PR #82, opened 2021-04-06
FIX use more reliable naming syntax for MKL version

### LINKED ISSUE #81: MKL warning (undefined symbol: mkl_get_version_string)
I was pointed here since a downstream project (fil profiler) said that this was a bug in threadpoolctl, but I think the library isn't able to handle our MKL version:

```
  File "_ctypes/callbacks.c", line 232, in 'calling callback function'
  File "<PATH>/lib/venv3/lib64/python3.7/site-packages/threadpoolctl.py", line 400, in match_module_callback
    self._make_module_from_path(filepath)
  File "<PATH>/lib/venv3/lib64/python3.7/site-packages/threadpoolctl.py", line 515, in _make_module_from_path
    module = module_class(filepath, prefix, user_api, internal_api)
  File "<PATH>/lib/venv3/lib64/python3.7/site-packages/threadpoolctl.py", line 606, in __init__
    self.version = self.get_version()
  File "<PATH>/lib/venv3/lib64/python3.7/site-packages/threadpoolctl.py", line 711, in get_version
    self._dynlib.mkl_get_version_string(res, 200)
  File "/opt/python/python-3.7/lib64/python3.7/ctypes/__init__.py", line 372, in __getattr__
    func = self.__getitem__(name)
  File "/opt/python/python-3.7/lib64/python3.7/ctypes/__init__.py", line 377, in __getitem__
    func = self._FuncPtr((name_or_ordinal, self))
AttributeError: /usr/local/lib/libmkl_rt.so: undefined symbol: mkl_get_version_string
```

I'm pretty sure the MKL version (as defined by `INTEL_MKL_VERSION`) is `20190004` if that helps.

jeremiedbb:
Fixes #81 

It seems that the naming syntax `MKL_My_Function` is safer to use than `mkl_my_function` which may not exist. We already observed that (for windows for instance iirc), this is why we already use the first syntax everywhere. I should have done that for the version as well in the first place.

Let's wait confirmation from the discussion in #81 before merging.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,10 @@
   (`openblas_get_corename`) in `threadpool_info()` and BLIS
   (`bli_arch_query_id` and `bli_arch_string`).
 
+- Fixed a bug when the version of MKL was not found. The
+  "version" field is now set to None in that case.
+  https://github.com/joblib/threadpoolctl/pull/82
+
 2.1.0 (2020-05-29)
 ==================
 
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -708,7 +708,7 @@ class _MKLModule(_Module):
     """Module class for MKL"""
     def get_version(self):
         res = ctypes.create_string_buffer(200)
-        self._dynlib.mkl_get_version_string(res, 200)
+        self._dynlib.MKL_Get_Version_String(res, 200)
 
         version = res.value.decode("utf-8")
         group = re.search(r"Version ([^ ]+) ", version)
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -707,6 +707,9 @@ class _BLISModule(_Module):
 class _MKLModule(_Module):
     """Module class for MKL"""
     def get_version(self):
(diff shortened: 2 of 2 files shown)


jeremiedbb:
Based on the last comments is the original issue, let's also check if the symbol exists, otherwise set the version to None (as we do when we can't retrieve the version for the other libs).

tomMoral (approved):
LGTM! I think doing a best effort and returning unknown if this is not working is probably the best way to go.

Maybe to be more robust, we should do a `try..except` on [l.606](https://github.com/joblib/threadpoolctl/blob/master/threadpoolctl.py#L606) and put `unknown(error)` when it fails?

jeremiedbb:
> Maybe to be more robust, we should do a try..except on l.606 and put unknown(error) when it fails?

But we could miss important errors that we can fix like this one. Also it would return unknown when it should actually work but we didn't have the occasion to see an error.

ogrisel:
I think this needs an entry in the changelog.

ogrisel (approved):
+1 with a changelog entry.

PR #83, opened 2021-04-15
[MRG] Fix compilation with clang and flake8 install in CI

jeremiedbb:
python introduced a compile and link flag that is not supported by old versions of clang (-fno-semantic-interposition).
installing clang 10 fixes it.

the installation of flake8 was not working (don't know why). I just changed the install command.


continuous_integration/install_with_blis.sh
```diff
@@ -7,11 +7,9 @@ ABS_PATH=$(pwd)
 popd
 
 # Assume Ubuntu: install a recent version of clang and libomp
```
ogrisel, line 9:
```suggestion
# Install a recent version of clang and libomp
```

ogrisel (approved):
LGTM, thanks for the fix!

PR #86, opened 2021-07-08
Update Python versions (3.6 min)

ogrisel:
Change CI and min version requirements in pyproject.toml.

### DIFF
--- a/.azure_pipeline.yml
+++ b/.azure_pipeline.yml
@@ -29,21 +29,21 @@ stages:
   - template: continuous_integration/posix.yml
     parameters:
       name: Linux
-      vmImage: ubuntu-18.04
+      vmImage: ubuntu-20.04
       matrix:
-        # Linux environment to test that packages that comes with Ubuntu 18.04
+        # Linux environment to test that packages that comes with Ubuntu 20.04
         # are correctly handled.
-        py36_ubuntu_atlas_gcc_gcc:
+        py38_ubuntu_atlas_gcc_gcc:
           PACKAGER: 'ubuntu'
           APT_BLAS: 'libatlas3-base libatlas-base-dev'
-          VERSION_PYTHON: '3.6'
+          VERSION_PYTHON: '3.8'
           CC_OUTER_LOOP: 'gcc'
           CC_INNER_LOOP: 'gcc'
           LINT: 'true'
         py36_ubuntu_openblas_gcc_gcc:
           PACKAGER: 'ubuntu'
           APT_BLAS: 'libopenblas-base libopenblas-dev'
-          VERSION_PYTHON: '3.6'
+          VERSION_PYTHON: '3.8'
           CC_OUTER_LOOP: 'gcc'
           CC_INNER_LOOP: 'gcc'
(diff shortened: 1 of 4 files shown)


ogrisel:
@jeremiedbb fbc08db "fixed" the segfault. This segfault happens with MKL and libomp from homebrew on macos...

This is bad but I am too lazy to investigate. And probably very few users will use that combo in practice.

ogrisel:
I tried to update the ubuntu version to 18.04 because 16.04 is deprecated on Azure Pipelines but now I get a weird error when importing scipy in the tests... It seems to be scipy problem but I am not sure what to do. Maybe we can try with ubuntu 20.04.

ogrisel:
Ok now the ubuntu packager builds work but all the builds that need the clang / libomp from the llvm installation script do not work anymore:

```
/usr/bin/ld: cannot find -lomp
```

ogrisel:
Ok hopefully the last commit will fix the CI 🤞

ogrisel:
This time it will work 🤞 🤞 🤞

ogrisel:
@jeremiedbb ready for merge. The segfault is not fixed, it's just hidden.

.azure_pipeline.yml
```diff
@@ -125,10 +125,10 @@ stages:
       vmImage: macOS-10.14
       matrix:
         # MacOS environment with OpenMP installed through homebrew
-        py35_conda_homebrew_libomp:
+        py36_conda_homebrew_libomp:
           PACKAGER: 'conda'
-          VERSION_PYTHON: '3.5'
-          BLAS: 'mkl'
+          VERSION_PYTHON: '3.6'
+          BLAS: 'openblas'
```
jeremiedbb, line 131: So the issue with mkl is not only on python 3.5 ?
It's a general issue with OpenMP installed from homebrew ? Because the setup just below seems to work with OpenMP installed from conda-forge.

continuous_integration/install.sh
```diff
@@ -9,6 +9,7 @@ if [[ "$CC_OUTER_LOOP" == "clang-10" || "$CC_INNER_LOOP" == "clang-10" ]]; then
     wget https://apt.llvm.org/llvm.sh
     chmod +x llvm.sh
     sudo ./llvm.sh 10
+    sudo apt-get install libomp-dev
```
jeremiedbb, line 12: I don't really understand this addition. Clang doesn't come with the LLVM OpenMP runtime ?

.azure_pipeline.yml
ogrisel: It's related to the Python version. It's probably dependent on the version of libomp and maybe MKL.

continuous_integration/install.sh
ogrisel: Not anymore apparently.

ogrisel: Or at least `llvm.sh` does not install this package by default anymore.

jeremiedbb:
ok LGTM, let's merge !

jeremiedbb (approved):

PR #87, opened 2021-07-20
Fix import with -OO

peterbell10:
When python is run with `-OO`, docstrings are removed and `__doc__` is set to `None`. This makes `_format_docstring` fail with the following error:
```
>>> import threadpoolctl
  File "<stdin>", line 1, in <module>
  File "/home/peter/miniconda3/lib/python3.8/site-packages/threadpoolctl.py", line 109, in <module>
    def threadpool_info():
  File "/home/peter/miniconda3/lib/python3.8/site-packages/threadpoolctl.py", line 101, in decorator
    o.__doc__ = o.__doc__.format(*args, **kwargs)
AttributeError: 'NoneType' object has no attribute 'format'
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@
   missing version query functions.
   https://github.com/joblib/threadpoolctl/pull/88
 
+- Fixed an attribute error when python is run with -OO.
+  https://github.com/joblib/threadpoolctl/pull/87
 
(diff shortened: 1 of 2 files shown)


jeremiedbb (approved):
LGTM. Thanks @peterbell10 !

PR #88, opened 2021-07-20
Fix getattr uses with invalid default values

peterbell10:
If the lambdas return None here, then `split` or `decode` will raise, e.g.
```
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.8.11/x64/lib/python3.8/site-packages/threadpoolctl.py", line 400, in match_module_callback
    self._make_module_from_path(filepath)
  File "/opt/hostedtoolcache/Python/3.8.11/x64/lib/python3.8/site-packages/threadpoolctl.py", line 515, in _make_module_from_path
    module = module_class(filepath, prefix, user_api, internal_api)
  File "/opt/hostedtoolcache/Python/3.8.11/x64/lib/python3.8/site-packages/threadpoolctl.py", line 606, in __init__
    self.version = self.get_version()
  File "/opt/hostedtoolcache/Python/3.8.11/x64/lib/python3.8/site-packages/threadpoolctl.py", line 646, in get_version
    config = get_config().split()
AttributeError: 'NoneType' object has no attribute 'split'

```

PR #91, opened 2021-09-02
Make threadpoolctl robust to missing openblas_get_parallel symbol

ogrisel:
This is a follow-up to #88.

I got this when testing against the openblas shipped with the [scipy nightly wheel for Apple M1](https://github.com/scipy/scipy/issues/13409#issuecomment-904548388).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 - Fixed an attribute error when using old versions of OpenBLAS or BLIS that are
   missing version query functions.
   https://github.com/joblib/threadpoolctl/pull/88
+  https://github.com/joblib/threadpoolctl/pull/91
 
 - Fixed an attribute error when python is run with -OO.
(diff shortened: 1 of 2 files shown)


jeremiedbb:
It's weird because it seems that it's the last version of OpenBLAS shipped in this wheel. Maybe they don't all symbols for this system yet.

The failure is because this PR is prior to the merge of the CI fix right ?

ogrisel:
> It's weird because it seems that it's the last version of OpenBLAS shipped in this wheel. Maybe they don't all symbols for this system yet.

I agree I do not understand. There is probably a problem with this OpenBLAS binary but it's still a good idea to make threadpoolctl more robust.

> The failure is because this PR is prior to the merge of the CI fix right ?

Probably. Let me merge.

ogrisel:
I merged by mistake... There was still a bug. Let me fix it.

PR #93, opened 2021-09-02
Fix typo in get_threading_layer

ogrisel:
I made a typo in #91, I am too tired...

### DIFF
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -671,7 +671,7 @@ class _OpenBLASModule(_Module):
                                         None)
         if openblas_get_parallel is None:
             return "unknown"
-        threading_layer = openblas_get_parallel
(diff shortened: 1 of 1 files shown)


ogrisel:
We don't have tests for this.

jeremiedbb:
It's not that easy to test since we don't build openblas (we build blis) and it can't be changed at runtime (like mkl) so it would be hard to find which threading_layer we expect to find unless we introspect the env to find how it was installed and maintain which packager ships openblas with which threading layer (although it might always be pthreads but not sure about that).

ogrisel:
We could have a smoke test that only checks that the output belongs to a list of valid possible results.

jeremiedbb:
Ok I'll try to come up with something

PR #95, opened 2021-09-16
[MRG] FEA Make it possible to pass pre-inspected modules to threadpool_limit

### LINKED ISSUE #94: Make it possible to pass pre-inspected modules to threadpool_limit
This would be useful for instance in scikit-learn to avoid re-introspecting the runtime each time we want to protect against over-subscription. Runtime introspection can be costly and it is often the case that once we have imported the libraries we are interested in, we do not expect the list of native threadpool modules to change.

More specifically, in the context of scikit-learn, we can assume that the runtimes (BLAS and OpenMP) will not change once we have imported scikit-learn (and recursively scipy to get the Cython level access to BLAS), so it would be more efficient to store the list of modules as a private attribute of the `sklearn` package and then reuse that list whenever we call `threadpool_limit` in scikit-learn estimators.

jeremiedbb:
Fixes #94 

In addition, makes some stuff public and get rid of the "module" naming, as suggested by @ogrisel
- `_ThreadpoolInfo` -> `ThreadpoolController`
- `_Module` -> `LibController`
- `_<Lib>Module` -> `<Lib>Controller`

The possibility to pass an existing controller (which has already loaded the libraries) to threadpool_limits makes it much faster (like 100x faster):
```py
In [1]: from threadpoolctl import ThreadpoolController, threadpool_limits

In [2]: controller = ThreadpoolController()

In [3]: %%timeit
   ...: with threadpool_limits(1):
   ...:     pass
   ...: 
190 µs ± 744 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [4]: %%timeit
   ...: with threadpool_limits(1, controller=controller):
   ...:     pass
   ...: 
1.43 µs ± 9.68 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

Although it's already quite short, it's not negligible in some situations (I already experienced that in scikit-learn).

TODO
- [x] add tests
- [x] finish doc
- [x] change log

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,10 +113,28 @@ that are loaded when importing Python packages:
   'version': None}]
 ```
 
-In the above example, `numpy` was installed from the default anaconda channel and
-comes with the MKL and its Intel OpenMP (`libiomp5`) implementation while
-`xgboost` was installed from pypi.org and links against GNU OpenMP (`libgomp`)
-so both OpenMP runtimes are loaded in the same Python program.
+In the above example, `numpy` was installed from the default anaconda channel and comes
+with MKL and its Intel OpenMP (`libiomp5`) implementation while `xgboost` was installed
+from pypi.org and links against GNU OpenMP (`libgomp`) so both OpenMP runtimes are
+loaded in the same Python program.
+
+The state of these libraries is also accessible through the object oriented API:
+
+```python
+>>> from threadpoolctl import threadpool_info
+>>> from pprint import pprint
+>>> import numpy
+>>> controller = ThreadpoolController()
+>>> pprint(controller.info())
+[{'architecture': 'Haswell',
+  'filepath': '/home/jeremie/miniconda/envs/dev/lib/libopenblasp-r0.3.17.so',
+  'internal_api': 'openblas',
+  'num_threads': 4,
+  'prefix': 'libopenblas',
+  'threading_layer': 'pthreads',
+  'user_api': 'blas',
+  'version': '0.3.17'}]
+```
 
 ### Setting the Maximum Size of Thread-Pools
 
@@ -124,16 +142,30 @@ Control the number of threads used by the underlying runtime libraries
 in specific sections of your Python program:
 
 ```python
-from threadpoolctl import threadpool_limits
-import numpy as np
+>>> from threadpoolctl import threadpool_limits
+>>> import numpy as np
 
+>>> with threadpool_limits(limits=1, user_api='blas'):
+...     # In this block, calls to blas implementation (like openblas or MKL)
+...     # will be limited to use only one thread. They can thus be used jointly
+...     # with thread-parallelism.
+...     a = np.random.randn(1000, 1000)
+...     a_squared = a @ a
+```
 
-with threadpool_limits(limits=1, user_api='blas'):
-    # In this block, calls to blas implementation (like openblas or MKL)
-    # will be limited to use only one thread. They can thus be used jointly
-    # with thread-parallelism.
-    a = np.random.randn(1000, 1000)
-    a_squared = a @ a
+The threadpools can also be controlled via the object oriented API, which is especially
+useful to avoid searching through all the loaded shared libraries each time. It will
+however not act on libraries loaded after the instanciation of the
+``ThreadpoolController``:
+
+```python
+>>> from threadpoolctl import ThreadpoolController
+>>> import numpy as np
+>>> controller = ThreadpoolController()
+
+>>> with controller.limit(limits=1, user_api='blas'):
+...     a = np.random.randn(1000, 1000)
+...     a_squared = a @ a
 ```
 
 ### Known Limitations
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ loaded in the same Python program.
 The state of these libraries is also accessible through the object oriented API:
 
 ```python
->>> from threadpoolctl import ThreadpoolController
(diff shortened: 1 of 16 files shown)


jeremiedbb:
@ogrisel, finally I chose to allow `threadpool_limits` to accept a controller at init. It felt weird to have a method of ThreadpoolController to just return `threadpool_limits(controller=self)`. The context manager holds a controller, not the opposite.
I'm still open to discussion and it's easy to add.

ogrisel:
From a user point of view, I would fined more natural to have:


```python

with controller.limit(1):
    # do stuff here
```

tests/test_threadpoolctl.py
```diff
@@ -14,10 +15,10 @@
 from .utils import threadpool_info_from_subprocess
 
 
-def is_old_openblas(module):
+def is_old_openblas(libctl):
```
ogrisel, line 18: Style: I think I would prefer the more explicit `lib_controller`.

ogrisel (commented):
To me `ctl` is a shortcut for "control" rather than "controller". And here the "old" modules are controllers, so I would explicitly name them as such.

While we are at it, it the longer `_controller` local variable names make it a hell to reformat to PEP8 manually, feel free to blackify everything while we are at it :)

threadpoolctl.py
```diff
@@ -774,24 +763,21 @@ def get_threading_layer(self):
         return layer_map[set_threading_layer(-1)]
 
 
-class _OpenMPModule(_Module):
-    """Module class for OpenMP"""
-    def get_version(self):
+class OpenMPController(LibController):
+    """Controller class for OpenMP"""
+    def _get_version(self):
```
ogrisel, line 768: I think we could keep the get/set methods public on the concrete controller objects.

tests/test_threadpoolctl.py
jeremiedbb: I was thinking the same but I thought it would make lines too long too often and hurt readability but actually it's fine

threadpoolctl.py
jeremiedbb: I don't mind

jeremiedbb:
I extracted the blackify part in another PR to distinguish between the 2 big source of diff.

ogrisel:
We can ignore the lint issues because they will be fixed in the black PR.

@jeremiedbb this PR is still WIP, do you still have changes to implement in mind, or is it ready for review?

threadpoolctl.py
```diff
@@ -105,36 +112,36 @@ def decorator(o):
...
+)
+def threadpool_limits(limits=None, user_api=None):
     """Change the maximal number of threads that can be used in thread pools.
 
-    This class can be used either as a function (the construction of this
-    object limits the number of threads) or as a context manager, in a `with`
-    block.
+    This function returns a class that can be used either as a function (the
+    construction of this object limits the number of threads) or as a context manager,
+    in a `with` block.
```
ogrisel, line 144:
```suggestion
    in a `with` block to automatically restore the original state of the
    controlled libraries when exiting the block.
```

ogrisel (commented):
Thank you very much for this PR. I like the design and naming.

In addition to the changelog, I think the new extended object oriented public API should be presented in a code snippet in the `README.md` of the project.

ogrisel:
For the record, I am fine with renaming `.todicts()` to `.infos()` if you wish @jeremiedbb.

README.md
```diff
@@ -113,27 +113,59 @@ that are loaded when importing Python packages:
...
+loaded in the same Python program.
+
+The state of these libraries is also accessible through the object oriented API:
+
+```python
+>>> from threadpoolctl import threadpool_info
+>>> from pprint import pprint
+>>> import numpy
+>>> controller = ThreadpoolController()
+>>> pprint(controller.info())
```
ogrisel, line 128: Here I would just use the imported function.

```suggestion
>>> pprint(threadpool_info())
```

README.md
```diff
@@ -113,27 +113,59 @@ that are loaded when importing Python packages:
...
 
-with threadpool_limits(limits=1, user_api='blas'):
-    # In this block, calls to blas implementation (like openblas or MKL)
-    # will be limited to use only one thread. They can thus be used jointly
-    # with thread-parallelism.
-    a = np.random.randn(1000, 1000)
-    a_squared = a @ a
+>>> with controller.limit(limits=1, user_api='blas'):
+...     a = np.random.randn(1000, 1000)
+...     a_squared = a @ a
```
ogrisel, line 168: Maybe you could add:

```suggestion
...     a_squared = a @ a
...
>>> controller.infos() == threadpool_info()
True
```

ogrisel (approved):
LGTM once the conflicts have been solved and the suggestions below have been handled.

README.md
jeremiedbb: I meant to import the class. There's an example with the function just above and this snippet is here to describe the object oriented API.

README.md
jeremiedbb: I put this check in the previous snippet where I introduce ThreadpoolController.info() to show their equivalence

ogrisel (approved):

ogrisel:
Merged! Thank you very much @jeremiedbb . Could you please try to leverage this in a PR in scikit-learn using a dummy backward compat wrapper in `sklearn.utils.fixes` to check that it works as expected? If so we can release threadpoolctl 3.0.0.

PR #96, opened 2021-09-17
MNT Blackify the code base

jeremiedbb:
Proposed in #95. Done in a separate PR to not interfere with the other changes.


jeremiedbb:
Seems good. I moved the lint job to `pylatest_conda_forge` for convenience.

ogrisel (approved):
LGTM but we could also add some config in `pyproject.toml` similarly to what we do in scikit-learn for instance:

```
[tool.black]
line-length = 88
target_version = ['py36', 'py37', 'py38', 'py39']
experimental_string_processing = true
```

to better control the behavior of black in case of changes.

jeremiedbb:
I added the config in pyproject.toml. Let's merge

PR #97, opened 2021-09-17
MNT use lru_cache for realpath

jeremiedbb:
An old comment says we should be using a lru_cache for the real paths when we drop python 2 support. I don't know if we ever supported it but our min python version is currently 3.6.

The benchmarks script ``benchmarks/bench_context_manager_overhead.py `` does not detect any performance change.
master: ``Overhead per call: 0.315 +/-0.071 ms``
this PR: ``Overhead per call: 0.309 +/-0.062 ms``
results are within each other error bounds

### DIFF
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -18,6 +18,7 @@ import textwrap
 import warnings
 from ctypes.util import find_library
 from abc import ABC, abstractmethod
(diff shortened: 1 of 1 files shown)


ogrisel (approved):
Nice clean-up, thanks!

PR #98, opened 2021-09-17
MNT use fstrings when possible

jeremiedbb:
Since we support python >= 3.6 we can use fstring. The only occurence left of .format is for docstring for which fstring are not ideal.
PR built on top of #96 for convenience. No need to review it before #96 is merged

### DIFF
--- a/benchmarks/bench_context_manager_overhead.py
+++ b/benchmarks/bench_context_manager_overhead.py
@@ -28,7 +28,6 @@ for _ in range(args.n_calls):
     timings.append(time.time() - t)
 
 print(
-    "Overhead per call: {:.3f} +/-{:.3f} ms".format(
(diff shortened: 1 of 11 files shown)


benchmarks/bench_context_manager_overhead.py
```diff
@@ -28,7 +28,6 @@
     timings.append(time.time() - t)
 
 print(
-    "Overhead per call: {:.3f} +/-{:.3f} ms".format(
-        mean(timings) * 1e3, stdev(timings) * 1e3
+    f"Overhead per call: {mean(timings) * 1e3:.3f} +/-{stdev(timings) * 1e3:.3f} ms"
     )
```
ogrisel, line 32:
```suggestion
```

ogrisel (approved):

PR #102, opened 2021-09-27
FEA Add possibility to use threadpool_limits as a decorator

### LINKED ISSUE #69: Make threadpool_limits into a decorator
We need to set threadpool_limits to basically every function in our package. I found that it would be much easier if threadpool_limits was also a decorator: we won't need to add another indent to our entire codebase, which is good from the code style point of view. 

I believe it could be done easily, by inheriting threadpool_manager from contextlib.ContextDecorator

https://docs.python.org/3/library/contextlib.html#using-a-context-manager-as-a-function-decorator

jeremiedbb:
Fixes #69 

Introduces `threadpool_limits.wrap(limits, user_api)` and `ThreadpoolController().wrap(limits, user_api)` that can be used as decorators to restrict their scope to the scope of the decorated function.

```py

@controller.wrap(limits=1, user_api="blas")
def my_func():
     # Single threaded blas inside the function
     X @ Y
```

or
```py
from threadpoolctl import threadpool_limits

@threadpool_limits.wrap(limits=1, user_api="blas")
def my_func():
     # Single threaded blas inside the function
     X @ Y
```

### DIFF
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -595,8 +595,10 @@ def test_threadpool_controller_as_decorator():
     controller = ThreadpoolController()
     original_info = controller.info()
 
+    if any(info["num_threads"] < 2 for info in original_info):
+        pytest.skip("Test requires at least 2 CPUs on host machine")
     if not controller.select(user_api="blas"):
-        pytest.skip(f"Requires a blas runtime.")
+        pytest.skip("Requires a blas runtime.")
 
     def check_blas_num_threads(expected_num_threads):
         blas_controller = ThreadpoolController().select(user_api="blas")
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -580,24 +580,21 @@ def test_threadpool_controller_as_decorator():
     if not controller.select(user_api="blas"):
         pytest.skip(f"Requires a blas runtime.")
 
-    @controller.wrap(limits=1, user_api="blas")
-    def func1():
+    def check_blas_num_threads(expected_num_threads):
         blas_controller = ThreadpoolController().select(user_api="blas")
         assert all(
-            lib_controller.num_threads == 1
+            lib_controller.num_threads == expected_num_threads
             for lib_controller in blas_controller.lib_controllers
         )
 
-        func2()
+    @controller.wrap(limits=1, user_api="blas")
+    def outer_func():
+        _check_blas_num_threads(1)
+        inner_func()
+        _check_blas_num_threads(1)
 
     @controller.wrap(limits=2, user_api="blas")
-    def func2():
-        blas_controller = ThreadpoolController().select(user_api="blas")
-        assert all(
-            lib_controller.num_threads == 2
-            for lib_controller in blas_controller.lib_controllers
-        )
-
-    func1()
+    def inner_func():
+        _check_blas_num_threads(2)
 
     assert controller.info() == ThreadpoolController().info()
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -551,3 +551,34 @@ def test_architecture():
         else:
             # Not supported for other libraries
             assert "architecture" not in lib_info
+
+
+def test_threadpool_controller_as_decorator():
+    # Check that using the decorator can be nested and is restricted to the scope of
+    # the decorated function.
+    controller = ThreadpoolController()
+
+    if not controller.select(user_api="blas"):
+        pytest.skip(f"Requires a blas runtime.")
+
+    @controller.wrap(limits=1, user_api="blas")
+    def func1():
+        blas_controller = ThreadpoolController().select(user_api="blas")
+        assert all(
+            lib_controller.num_threads == 1
+            for lib_controller in blas_controller.lib_controllers
+        )
+
+        func2()
+
+    @controller.wrap(limits=2, user_api="blas")
+    def func2():
+        blas_controller = ThreadpoolController().select(user_api="blas")
+        assert all(
+            lib_controller.num_threads == 2
+            for lib_controller in blas_controller.lib_controllers
+        )
+
+    func1()
+
+    assert controller.info() == ThreadpoolController().info()
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -576,6 +576,7 @@ def test_threadpool_controller_as_decorator():
     # Check that using the decorator can be nested and is restricted to the scope of
     # the decorated function.
     controller = ThreadpoolController()
+    original_info = controller.info()
 
     if not controller.select(user_api="blas"):
         pytest.skip(f"Requires a blas runtime.")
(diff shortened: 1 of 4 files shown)


threadpoolctl.py
```diff
@@ -199,11 +155,29 @@ def __init__(self, controller, *, limits=None, user_api=None):
...
 
+    @classmethod
+    def wrap(cls, controller, *, limits=None, user_api=None):
+        """Return an instance of this class that can be used as a decorator"""
+        obj = type(cls.__name__, (ContextDecorator,), dict(cls.__dict__))
+
+        # we need to create an instance without calling __init__ because we want the
+        # limits to be set when calling the decorated function, not when creating the
+        # decorator.
+        new_ctx = obj.__new__(obj)
```
ogrisel, line 174: Wouldn't it be possible to call the constructor without limits and have a private method to update/set the limits after the object has been instantiated without hacking around with `obj.__new__`?

ogrisel (commented):
Not sure if there is not a less hacky way to achieve this but I like the resulting public API and test.

Please also document the new decorator API in the README.

threadpoolctl.py
jeremiedbb: I don't think so because when it's used as a callable we only create the instance, so the limits have to be set in the init.

jeremiedbb: I just thought of an alternative but I'm not sure it's less hacky :)
I'll push it and can see if it's better or not

jeremiedbb: So in this alternative implem, the wrap method of _threadpool_limits returns an instance of a child class of this class.

I thought there would be some kind of reference cycle contradiction, but python :)

Do you prefer this solution ?

ogrisel (commented):
I think I prefer this design. Could you please expand the readme to introduce the decorators? Maybe in a new paragraph/section?

ogrisel:
@mimakaev could you please tell us if this PR would solve your initial needs?

threadpoolctl.py
```diff
@@ -315,6 +279,76 @@ def _set_threadpool_limits(self):
...
+
+    This object can be used either as a callable (the construction of this object
+    limits the number of threads), as a context manager in a `with` block to
+    automatically restore the original state of the controlled libraries when exiting
+    the block, or as a decorator through its `wrap` method.
+
+    Set the maximal number of threads that can be used in thread pools used in
+    the supported libraries to `limit`. This function works for libraries that
+    are already loaded in the interpreter and can be changed dynamically.
+
```
ogrisel, line 315: Maybe we could add a disclaimer to the docstrings of the public method to make it explicit that even when limits are set using the context manager or function decorator APIs, the effect is global on whole Python process: the is no thread level isolation as the underlying libraries do not offer thread-local APIs to configure the number to use in nested parallel calls.

See: https://github.com/joblib/threadpoolctl/issues/69#issuecomment-707136187

jeremiedbb: The issue in the notebook from this comment is that the limits were set in the __init__ so creating the decorator would set the limits right away for all further blas calls.

But I agree that we can add that disclaimer anyway

tomMoral (commented):
The new API is very neat and the naming also makes more sense!
Just a small comment reading by the tests.

tests/test_threadpoolctl.py
```diff
@@ -570,3 +570,34 @@ def test_openblas_threading_layer():
...
+
+        func2()
+
+    @controller.wrap(limits=2, user_api="blas")
+    def func2():
+        blas_controller = ThreadpoolController().select(user_api="blas")
+        assert all(
+            lib_controller.num_threads == 2
+            for lib_controller in blas_controller.lib_controllers
+        )
```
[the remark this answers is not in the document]
jeremiedbb: thanks. I made it even more explicit by naming the functions inner_func and outer_func.

threadpoolctl.py
jeremiedbb: Done

threadpoolctl.py
```diff
@@ -138,51 +139,6 @@ def threadpool_info():
...
-        - If "openmp", it will only limit OpenMP supported libraries
-          ({OPENMP_LIBS}). Note that it can affect the number of threads used
-          by the BLAS libraries if they rely on OpenMP.
-
-        - If None, this function will apply to all supported libraries.
-    """
-    return ThreadpoolController().limit(limits=limits, user_api=user_api)
-
-
 class _threadpool_limits:
```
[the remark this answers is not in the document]
jeremiedbb: Done

tests/test_threadpoolctl.py
```diff
@@ -580,24 +580,21 @@ def test_threadpool_controller_as_decorator():
     if not controller.select(user_api="blas"):
         pytest.skip(f"Requires a blas runtime.")
 
-    @controller.wrap(limits=1, user_api="blas")
-    def func1():
+    def check_blas_num_threads(expected_num_threads):
```
tomMoral, line 583:
```suggestion
    def _check_blas_num_threads(expected_num_threads):
```

tests/test_threadpoolctl.py
```diff
@@ -580,24 +580,21 @@ def test_threadpool_controller_as_decorator():
...
-    def func2():
-        blas_controller = ThreadpoolController().select(user_api="blas")
-        assert all(
-            lib_controller.num_threads == 2
-            for lib_controller in blas_controller.lib_controllers
-        )
-
-    func1()
+    def inner_func():
+        _check_blas_num_threads(2)
```
tomMoral, line 599:
```suggestion

outer_func()

```

mimakaev:
this looks amazing! Thank you guys!

jeremiedbb:
Thanks to @tomMoral refactoring of the test, I found a bug, already present in the context manager (from the OO API, not from threadpool_limits), and opened a PR to fix it https://github.com/joblib/threadpoolctl/pull/104.

jeremiedbb:
#104 has been merged. I adapted the fix for the decorator. Everything looks fine now. Let's merge ?

tests/test_threadpoolctl.py
```diff
@@ -587,3 +587,34 @@ def test_openblas_threading_layer():
...
+
+
+def test_threadpool_controller_as_decorator():
+    # Check that using the decorator can be nested and is restricted to the scope of
+    # the decorated function.
+    controller = ThreadpoolController()
+    original_info = controller.info()
+
+    if not controller.select(user_api="blas"):
+        pytest.skip(f"Requires a blas runtime.")
```
tomMoral, line 599: Should we also skip if there is less than 2 cpus like in the nested case?
```suggestion
    blas_ctls = controller.select(user_api="blas")
    if not blas_ctls or any(ctl.num_threads < 2 for ctl in blas_ctls):
        pytest.skip(f"Requires a blas runtime and more than 2 CPUs.")
```

tomMoral (approved):
A small comment but otherwise LGTM. The aPI is very neat, thx @ogrisel  and @jeremiedbb .

tests/test_threadpoolctl.py
jeremiedbb: yep makes sense. Done (I split the 2 skipping options to know exactly why it was skipped)

PR #103, opened 2021-09-27
TST Add a test for openblas threading layer

jeremiedbb:
follow up of this comment https://github.com/joblib/threadpoolctl/pull/93#issuecomment-926588659

We can't test the real threading_layer because it would imply to introspect the conda env, but we can check that we find a value that is in the list of acceptable values.


ogrisel (approved):
That's a good way to test it :)

PR #104, opened 2021-09-29
FIX make the num_threads attribute dynamic

jeremiedbb:
Found a bug while working on #102 when nesting `ThreadpoolController.limit` context managers.

Since the controller holds the orignal num_threads from when it was created, restoring the limits when exiting the inner context sets the limits back to the original value from before the outer context instead of from within the outer context.
For that we need to retrieve and store the current state of the threadpools when entering the context.
It causes a slowdown from ~2us to ~10us, but we don't have a choice and it's still much faster that the ~300us of `threadpool_limits`

This bug made me also think that having the `num_threads` attribute being static is confusing and might lead to unexpected behavior. for instance:
```py
controller.info()  # original_info, e.g. num_threads = 4

threadpool_limits(limits=1)
controller.info() # still num_threads = 4
```

I propose to make the `num_threads` attribute of the library controllers dynamic, by calling `get_num_threads` each time we want to access it, such that it reflects the current state of the threadpools.

This change requires a lot of rewriting in the test because it used to rely a lot on the fact that the controller holds the original num_threads. Now we store the original info from `controller.info()` and use it to check that we restore the limits to their correct value.

### DIFF
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -13,6 +13,7 @@ from .utils import cython_extensions_compiled
 from .utils import libopenblas_paths
 from .utils import scipy
 from .utils import threadpool_info_from_subprocess
+from .utils import select
 
 
 def is_old_openblas(lib_controller):
@@ -60,6 +61,20 @@ def test_threadpool_controller_info():
             assert "threading_layer" in lib_controller_dict
 
 
+def test_controller_info_actualized():
+    # Check that the num_threads attribute reflects the actual state of the threadpools
+    controller = ThreadpoolController()
+    original_info = controller.info()
+
+    with threadpool_limits(limits=1):
+        assert all(
+            lib_controller.num_threads == 1
+            for lib_controller in controller.lib_controllers
+        )
+
+    assert controller.info() == original_info
+
+
 @pytest.mark.parametrize(
     "kwargs",
     [
@@ -86,9 +101,10 @@ def test_threadpool_controller_select(kwargs):
 @pytest.mark.parametrize("limit", [1, 3])
 def test_threadpool_limits_by_prefix(prefix, limit):
     # Check that the maximum number of threads can be set by prefix
-    original_controller = ThreadpoolController()
+    controller = ThreadpoolController()
+    original_info = controller.info()
 
-    controller_matching_prefix = original_controller.select(prefix=prefix)
+    controller_matching_prefix = controller.select(prefix=prefix)
     if not controller_matching_prefix:
         pytest.skip(f"Requires {prefix} runtime")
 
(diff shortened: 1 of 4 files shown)


tests/test_threadpoolctl.py
```diff
@@ -172,46 +190,49 @@ def test_threadpool_limits_manual_unregister():
...
 
-def test_threadpool_controller_restore():
-    # Check that the restore_limits method of ThreadpoolController is able to set the
-    # limits back to their original values. Similar to
-    # test_threadpool_limits_function_with_side_effect but with the object api
+def test_nested_limits():
+    # Check that exiting the context manager properly restores the original limits even
+    # when nested.
     controller = ThreadpoolController()
+    original_info = controller.info()
```
ogrisel, line 219: I think we should protect this test to avoid failing in case the host machine has a single CPU. Something in the spirit of the following untested snippet:

```suggestion
    original_info = controller.info()
    if any(info["num_threads"] < 2 for info in original_info):
        pytest.skip("Test requires at least 2 CPUs on host machine")
```

threadpoolctl.py
```diff
@@ -205,10 +206,11 @@ def __exit__(self, type, value, traceback):
         self.unregister()
 
     def unregister(self):
```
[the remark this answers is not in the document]
ogrisel: Beyond the naming of this method, I think most of (all?) the methods of the `_threadpool_limits` class should be marked private. Just in case people start with write code such as:

```python
with threadpool_limits(1) as ctx:
    ctx.unregister()
```

threadpoolctl.py
```diff
@@ -205,10 +206,11 @@ def __exit__(self, type, value, traceback):
...
-            # Since we never call get_num_threads after instanciation of
-            # ThreadpoolController, num_threads holds the original value.
-            lib_controller.set_num_threads(lib_controller.num_threads)
+        """Set the limits back to their original values"""
+        for lib_controller, original_info in zip(
+            self._controller.lib_controllers, self._original_info
+        ):
+            lib_controller.set_num_threads(original_info["num_threads"])
 
     def get_original_num_threads(self):
```
ogrisel, line 215: I wouldn't mind if this would stay public in the context of the previous comment.

ogrisel (approved):
Thanks for the PR. Overall, LGTM. See some suggestions below:

threadpoolctl.py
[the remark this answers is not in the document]
tomMoral: If all methods are private, why not simply return None in `__enter__`? this way it is not possible to use the `as ctx`?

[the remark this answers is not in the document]
jeremiedbb: as discussed irl with @ogrisel, they are already part of the public API so we'll keep them public. In addition `unregister` is useful to restore the limits when it was used as a callable.

It's useful to have access to the object inside the context, at least for testing purpose. I'd rather keep that.

tests/test_threadpoolctl.py
jeremiedbb: I suspect it's unreachable code in our CI, where I think we always have at least 2 cores. But better too safe than not enough

threadpoolctl.py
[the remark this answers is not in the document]
ogrisel: Alright. Let's keep the existing method public with an alias with a better name for unregister but no need to deprecate the old name.

threadpoolctl.py
ogrisel: But no strong opinion either on this one?

jeremiedbb: Let's keep it for backward compat ?

threadpoolctl.py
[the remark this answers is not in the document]
tomMoral: Ok!

tomMoral (approved):
LGTM!

threadpoolctl.py
```diff
@@ -196,19 +196,25 @@ def __init__(self, controller, *, limits=None, user_api=None):
...
+        ):
+            lib_controller.set_num_threads(original_info["num_threads"])
 
     def unregister(self):
-        for lib_controller in self._controller.lib_controllers:
-            # Since we never call get_num_threads after instanciation of
-            # ThreadpoolController, num_threads holds the original value.
-            lib_controller.set_num_threads(lib_controller.num_threads)
+        """Alias of `restore_original_limits` for backward compatibility"""
+        self.restore_original_limits()
```
ogrisel, line 217: I wonder if the following wouldn't work:

```python
    # Backward compat alias
    unregister = restore_original_limits
```

ogrisel: I tested it locally and it should work.

jeremiedbb: it does indeed. thanks

jeremiedbb:
All green + All comments resolved. Let's merge. Thanks for the reviews

PR #112, opened 2021-12-31
MNT CI updates

jeremiedbb:


threadpoolctl.py
```diff
@@ -660,6 +667,19 @@ def _make_controller_from_path(self, filepath):
             if prefix is None:
                 continue
 
+            # workaround for blas install from conda-forge on windows. They rename all
+            # blas libs into "libblas.dll". We then have to check to which blas
```
ogrisel, line 671:
```suggestion
            # blas libs to "libblas.dll". We then have to check to which blas
```

threadpoolctl.py
```diff
@@ -660,6 +667,19 @@ def _make_controller_from_path(self, filepath):
...
+            # implementation it corresponds.
+            if prefix == "libblas":
+                if filename.endswith(".dll"):
```
ogrisel, line 674: I wonder if we should stay windows-specific or try to be blindly future-proof and do this for any platform. I think I would favor the later which should also be simpler from a code point of view. WDYT?

ogrisel (approved):
LGTM. The PR title should be update to reflect the change of scope of this PR.

threadpoolctl.py
jeremiedbb: I had to make it windows specific because there was a "libblas.so.3" on linux that came from "/usr/lib/x86_64-linux-gnu/openblas-pthread/libblas.so.3" that does not expose all the symbols we want. I don't know what it is exactly. See https://dev.azure.com/joblib/threadpoolctl/_build/results?buildId=1395&view=logs&j=49602bd8-c265-5cf5-ff6d-9ba0d4e0f8a9&t=7b75b0ca-af55-5daa-918c-61fa8f39fc20

ogrisel: Maybe add an inline comment for this.

ogrisel: Why point was more to mention that ubuntu (and maybe debian) also includes a `libblas` as part of its openblas distribution with some but not all the openblas specific symbols in addition to a `libopenblas` that comes with all symbols.

threadpoolctl.py
```diff
@@ -73,28 +75,33 @@ class _dl_phdr_info(ctypes.Structure):
     "OpenBLASController": {
         "user_api": "blas",
         "internal_api": "openblas",
-        "filename_prefixes": ("libopenblas",),
+        "filename_prefixes": ("libopenblas", "libblas"),
+        "check_funcs": ("openblas_get_num_threads", "openblas_get_num_threads64_"),
```
ogrisel, line 79: Maybe `check_symbols` would make it easier to understand the purpose of this key in the spec?

ogrisel:
Thanks for the fix!

PR #114, opened 2022-01-05
Workaround for openblas + openmp threading layer

jeremiedbb:


ogrisel:
I think we need a dedicated API to protect about the `sequential_blas_under_openmp` oversubscription case that we need in scikit-learn.

- For an  OpenBLAS BLAS implementation with an OpenMP threading layer and if there is a single OpenMP runtime loaded, assume that OpenMP will already handle the case natively and do nothing;
- For an  OpenBLAS implementation with a threading layer that is either native or OpenMP but there are several active OpenMP runtimes linked to the Python process, set `limit=1` on this OpenBLAS;
- For MKL and BLIS with OpenMP threading layers, we need to check.

This should be a new context manager, both for the functional and the pre-loaded object oriented API.

WDYT @jeremiedbb?

jeremiedbb:
So at first I tried this new context manager idea but I was very unsatisfied with it. I think I came up with a more satisfying solution. When we set the limit to blas only, we check if the internal state of openmp has changed and if so we set it back to the original state (this does not modify the state of the blas, I checked).

I thinks it's better to just have this kind of workaround for a behavior that is unexpected at least and should even be considered a bug imo, than adding a new whole public object just to handle that. What do you think ?

jeremiedbb:
Hum, actually setting back openmp to its original state is bad when we don't want to set the limits for the nested openmp openblas case... It will run as if no limit was set even if `openblas_get_num_threads` returns 1.

jeremiedbb:
Ok so I think I came up with an acceptable alternative: a new helper function that gives the appropriate params to pass to `threadpool_limits` only for the use case of nested blas inside an openmp parallel section. There's no need to introduce a new context manager.

ogrisel:
Nice!

Maybe this could also be somehow exposed in the `.limit` function? For instance

```python
with threadpoolctl.threadpool_limits(limits="sequential_blas_under_openmp"):
    # code protected against over-subscription
```

WDYT @jeremiedbb?

If you agree, please update the README to document that high level policy.

.azure_pipeline.yml
```diff
@@ -81,7 +81,7 @@ stages:
           VERSION_PYTHON: '3.8'
           CC_OUTER_LOOP: 'gcc'
           CC_INNER_LOOP: 'clang-10'
-        # Linux environment with numpy from conda-forge channel
+        # Linux environment with numpy from conda-forge channel and openblas-openmp
```
ogrisel, line 84: I think the configuration needs to be updated to enforce the openmp threading layer for openblas  because the default is platform specific and might depend on the version of the package.

continuous_integration/install.sh
```diff
@@ -50,6 +50,9 @@ elif [[ "$PACKAGER" == "conda-forge" ]]; then
     conda config --prepend channels conda-forge
     conda config --set channel_priority strict
     TO_INSTALL="python=$VERSION_PYTHON numpy scipy blas[build=$BLAS]"
+    if [[ "$BLAS" == "openblas" ]]; then
+        TO_INSTALL="$TO_INSTALL libopenblas=*=*openmp*"
+    fi
```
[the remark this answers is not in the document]
jeremiedbb: I agree. I added a separate env var for that.

jeremiedbb:
> Maybe this could also be somehow exposed in the .limit function?

I like this Idea. Actually I think it should be the only public way to do that and I would make the helper private. Ok for you ?

ogrisel:
I agree. Let's make the getter private. The changelog will need to be updated accordingly to only mention the new limit option.

ogrisel (approved):
LGTM!

ogrisel:
This PR is missing a changelog entry though.

CHANGES.md
```diff
@@ -4,6 +4,13 @@
 - Fixed a detection issue of the BLAS libraires packaged by conda-forge on Windows.
   https://github.com/joblib/threadpoolctl/pull/112
 
+- `threadpool_info` and `ThreadpoolController.limit` accept a new value for the `limits`
+  parameter: the string "sequential_blas_under_openmp". It should only be used for the
```
ogrisel, line 8:
```suggestion
- `threadpool_info` and `ThreadpoolController.limit` now accept the string
  "sequential_blas_under_openmp" for the `limits` parameter.
  It should only be used for the
```

ogrisel (approved):
I find the construction of the first sentence in the changelog entry a bit weird. See below:

CHANGES.md
jeremiedbb: yeah I was not happy with my phrasing. Thanks

ogrisel:
Still `[WIP]`? If not, +1 for merging and maybe Friday evening release ;) Or maybe not releasing today.

CHANGES.md
```diff
@@ -4,6 +4,13 @@
 - Fixed a detection issue of the BLAS libraires packaged by conda-forge on Windows.
   https://github.com/joblib/threadpoolctl/pull/112
 
+- `threadpool_info` and `ThreadpoolController.limit` now accept the string
```
ogrisel, line 7:
```suggestion
- `threadpool_limit` and `ThreadpoolController.limit` now accept the string
```

jeremiedbb:
> maybe Friday evening release

I can't think of a better time to do this. But unfortunately I don't have time and I'm still not crazy enough to make a half release a friday evening and finish it during the week-end :D

jeremiedbb:
Ok all green. Let's merge

PR #133, opened 2023-02-15
MAINT whitelist test_multiple_openblas for check_no_test_always_skipped

jeremiedbb:
`test_multiple_openmp` checks that we end up with 2 different openblas with numpy and scipy. However it can happen that all the CI jobs have versions of these 2 libs that are in sync with their openblas versions.

We don't want to maintain a pin on specific versions of numpy and scipy only to trigger this situation. Instead, we'd rather just ignore this test in the check. If at some point the situation happens again the test will still be here and will hopefully still pass :)


ogrisel (approved):

ogrisel:
Thanks!

Issue #137, opened 2023-06-29
Support registering user-defined `LibController`


alugowski:
Very useful library.

I've gotten a request for my own library to be able to be controlled via `threadpoolctl`, and I agree that would be a positive thing to support, but it's currently impossible. While `threadpoolctl` is well designed and it appears to be easy to support new parallel libraries, the list of `LibController`s is static.

Would it be possible to add a feature where libraries could add support to register themselves with `threadpoolctl`?

For example, something like
```python
try:
    import threadpoolctl
    threadpoolctl.register(MyLibController)
except ImportError:
    pass
```

I imagine it's a big burden for the `threadpoolctl` authors to support all popular libraries out there, but with such a feature those library authors (or their users) could write such hooks ourselves.

There are several open requests for [Apple Accelerate](https://github.com/joblib/threadpoolctl/issues/135), [arrow](https://github.com/joblib/threadpoolctl/issues/134), [TensorFlow](https://github.com/joblib/threadpoolctl/issues/84), and I would be interested in taskflow or Eigen's threadpool, too. Additionally there are many smaller libraries, or ones using Python threads, that could easily be supported by bespoke `LibController`s with no effort on your part.

tomMoral:
I think this would be interesting to have this feature, as this would make it simpler to support various libs compared to having a centralized implems here.

If anyone wanna give it a try, feel free to propose a PR.

jeremiedbb:
I also find this approach interesting for libraries that we're not likely to support directly. I'm checking if it would be easy to implement.

ogrisel:
@alugowski #138 was merged with a fix. Please let us know if that good enough for your use case.

alugowski:
@ogrisel It *should* work, but my controller methods do not get called at all.

I installed via `pip install --upgrade git+https://github.com/joblib/threadpoolctl`

My controller code: https://github.com/alugowski/fast_matrix_market/blob/954331f9d65e4ce8b56dc8009cd09cc70e93f08e/python/src/fast_matrix_market/__init__.py#L49

and I call it like this:
```python
>>> from fast_matrix_market import mmread
registered with threadpoolctl
>>>
>>> from threadpoolctl import threadpool_limits
>>>
>>> with threadpool_limits(limits=1, user_api='fast_matrix_market'):
...    mmread("asdf.mtx")
...
parallelism: 0
```

ogrisel:
[...]

PR #138, opened 2023-07-05
FEA Support custom lib controllers

jeremiedbb:
Fixes https://github.com/joblib/threadpoolctl/issues/137

The goal is to expose the `LibController` abstract class, and custom controllers should inherit from it. This PR also exposes a new function `register` that simply adds the custom controller to the list of controllers to try.

TODO:
- [x] documentation

### DIFF
--- a/tests/_pyMylib/__init__.py
+++ b/tests/_pyMylib/__init__.py
@@ -14,13 +14,13 @@ class MyThreadedLibController(LibController):
     filename_prefixes = ("my_threaded_lib",)
 
     def get_num_threads(self):
-        return getattr(self._dynlib, "mylib_get_num_threads")()
+        return getattr(self.dynlib, "mylib_get_num_threads")()
 
     def set_num_threads(self, num_threads):
-        getattr(self._dynlib, "mylib_set_num_threads")(num_threads)
+        getattr(self.dynlib, "mylib_set_num_threads")(num_threads)
 
     def get_version(self):
-        get_version = getattr(self._dynlib, "mylib_get_version")
+        get_version = getattr(self.dynlib, "mylib_get_version")
         get_version.restype = ctypes.c_char_p
         return get_version().decode("utf-8")
 
--- a/tests/_pyMylib/__init__.py
+++ b/tests/_pyMylib/__init__.py
@@ -9,22 +9,37 @@ ctypes.CDLL(path)
 
 
 class MyThreadedLibController(LibController):
+    # names for threadpoolctl's context filtering
     user_api = "my_threaded_lib"
     internal_api = "my_threaded_lib"
+
+    # Patterns to identify the name of the linked library to load.
+    # If a dynamic library with a matching filename is linked to the python
+    # process, it will be loaded as the `dynlib` attribute of the LibController
+    # instance.
     filename_prefixes = ("my_threaded_lib",)
 
     def get_num_threads(self):
+        # This function should return the current maximum number of threads,
+        # which is reported as "num_threads" by `ThreadpoolController.info`.
         return getattr(self.dynlib, "mylib_get_num_threads")()
 
     def set_num_threads(self, num_threads):
+        # This function limits the maximum number of threads,
+        # when `ThreadpoolController.limit` is called.
         getattr(self.dynlib, "mylib_set_num_threads")(num_threads)
 
     def get_version(self):
+        # This function returns the version of the linked library if it is exposed,
+        # which is reported as "version" by `ThreadpoolController.info`.
         get_version = getattr(self.dynlib, "mylib_get_version")
         get_version.restype = ctypes.c_char_p
         return get_version().decode("utf-8")
 
     def set_additional_attributes(self):
+        # This function is called during the initialization of the LibController.
+        # Additional information meant to be exposed by `ThreadpoolController.info`
+        # should be set here as attributes of the LibController instance.
         self.some_attr = "some_value"
 
 
--- a/tests/_pyMylib/__init__.py
+++ b/tests/_pyMylib/__init__.py
@@ -1,0 +1,28 @@
+import ctypes
+from pathlib import Path
+
+from threadpoolctl import LibController, register
+
+
+path = Path(__file__).parent / "my_threaded_lib.so"
+ctypes.CDLL(path)
+
+
+class MyThreadedLibController(LibController):
+    user_api = "my_threaded_lib"
+    internal_api = "my_threaded_lib"
+    filename_prefixes = ("my_threaded_lib",)
+
+    def get_num_threads(self):
+        return getattr(self._dynlib, "mylib_get_num_threads")()
+
+    def set_num_threads(self, num_threads):
+        getattr(self._dynlib, "mylib_set_num_threads")(num_threads)
+
+    def get_version(self):
+        get_version = getattr(self._dynlib, "mylib_get_version")
+        get_version.restype = ctypes.c_char_p
+        return get_version().decode("utf-8")
+
+
+register(MyThreadedLibController)
--- a/tests/_pyMylib/__init__.py
+++ b/tests/_pyMylib/__init__.py
@@ -24,5 +24,8 @@ class MyThreadedLibController(LibController):
         get_version.restype = ctypes.c_char_p
         return get_version().decode("utf-8")
 
+    def set_additional_attributes(self):
+        self.some_attr = "some_value"
+
 
 register(MyThreadedLibController)
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -219,6 +219,7 @@ class BLISController(LibController):
     user_api = "blas"
     internal_api = "blis"
     filename_prefixes = ("libblis", "libblas")
+    check_symbols = ("bli_thread_get_num_threads",)
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -73,9 +73,10 @@ class LibController(ABC):
     A library controller must expose the following class attributes:
         - user_api : str
             Usually the name of the library or generic specification the library
-            implements.
+            implements, e.g. "blas" is a specification with different implementations.
         - internal_api : str
-            Usually the name of the library, or concrete implementation of some
+            Usually the name of the library or concrete implementation of some
+            specification, e.g. "openblas" is an implementation of the "blas"
             specification.
         - filename_prefixes : tuple
             Possible prefixes of the shared library's filename that allow to
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -15,6 +15,7 @@ import re
 import sys
 import ctypes
 import textwrap
+from typing import final
 import warnings
 from ctypes.util import find_library
 from abc import ABC, abstractmethod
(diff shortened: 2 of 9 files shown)


threadpoolctl.py
```diff
@@ -60,56 +66,306 @@ class _dl_phdr_info(ctypes.Structure):
...
-        "filename_prefixes": ("libmkl_rt", "mkl_rt", "libblas"),
-        "check_symbols": ("MKL_Get_Max_Threads",),
-    },
-    "BLISController": {
-        "user_api": "blas",
-        "internal_api": "blis",
-        "filename_prefixes": ("libblis", "libblas"),
-        "check_symbols": ("bli_thread_get_num_threads",),
-    },
-}
```
jeremiedbb, line 93: I was not happy that registering a now controller would require to pass a dict like that alongside the controller class. Therefore I refactored things a bit to make these entries class attributes of the lib controllers, which I find better overall because it allows to get rid of the awful
``lib_controller_class = globals()[controller_class]``.

tomMoral: Amazing ;)

threadpoolctl.py
```diff
@@ -60,56 +67,322 @@ class _dl_phdr_info(ctypes.Structure):
...
+    def internal_api(self):
+        """Internal API of the library"""
+
+    @property
+    @abstractmethod
+    def filename_prefixes(self):
+        """Possible prefixes of the library's filename"""
+
+    @property
+    @final
```
tomMoral, line 137: I did not know this one but this looks super useful!

threadpoolctl.py
```diff
@@ -60,56 +67,322 @@ class _dl_phdr_info(ctypes.Structure):
...
-        "check_symbols": ("bli_thread_get_num_threads",),
-    },
-}
+class LibController(ABC):
+    """Abstract base class for the individual library controllers
+
+    A library controller must expose the following class attributes:
+        - user_api : str
+            Usually the name of the library or generic specification the library
+            implements.
```
tomMoral, line 76: Maybe explain why it is used to give hints on how to select them?
```suggestion
            implements. This allows limiting generic frameworks
            regardless of the specific implementations, e.g. openMP.
```

I am actually not sure this one should be mandatory, maybe it could default to `internal_api`, or the other way around?

threadpoolctl.py
```diff
@@ -60,56 +67,322 @@ class _dl_phdr_info(ctypes.Structure):
...
+            identify the library. e.g. "libopenblas" for libopenblas.so.
+
+    A library contoller must implement the following methods: `get_num_threads`,
+    `set_num_threads` and `get_version`.
+
+    The following information will be exposed in the info dict:
+      - user_api : user API.
+      - internal_api : internal API.
+      - num_threads : the current thread limit.
+      - prefix : prefix of the shared library's filename.
```
tomMoral, line 91: To match the order they are implemented in:
```suggestion
      - prefix : prefix of the shared library's filename.
      - num_threads : the current thread limit.
```

threadpoolctl.py
```diff
@@ -60,56 +67,322 @@ class _dl_phdr_info(ctypes.Structure):
...
+      - filepath : path to the loaded shared library.
+      - version : version of the library (if available).
```
tomMoral, line 93: As this is not always possible to retrieve, is it necessary to make the `get_version` method mandatory?

threadpoolctl.py
```diff
@@ -60,56 +67,322 @@ class _dl_phdr_info(ctypes.Structure):
...
+
+    In addition, each library controller may expose internal API specific entries. They
+    must be set as attributes in the `set_additional_attributes method`.
+    """
```
tomMoral, line 97: Maybe something missing in the doc is to explain that we will loop through the linked libs to find libs matching the `filename_prefix` and that these libs will be available as `_dynlib`?
Otherwise, you need to know the internals of the API pretty well to be able to implement this

tomMoral (approved):
A few comment on the doc, but the implem is super clean!!
I like this very much :)

Thanks @jeremiedbb

threadpoolctl.py
jeremiedbb: I learned about it while making this PR :)
It's from typing though, not from abc, meaning it's only used by type checkers but there's no runtime error if you still override a method decorated as final.

threadpoolctl.py
jeremiedbb: The current order is their oder in the info dict returned by ``threadpool_info`` which makes sense as well.

threadpoolctl.py
jeremiedbb: I'd rather do that. If it can't be retreived, this method can return None or "not found" or whatever, so it's not a blocker. The thing is I know peole are lazy and they won't implement it if we don't make it mandatory, which is bad because the version info is usually very important when debugging :)
Also, having to fallback when the version can't be retreive might make people think that it's probably a good thing to have a programmatic way to find the version number of the library, so we're encouraging good practices :smile:

threadpoolctl.py
jeremiedbb: I was actually in the middle of making it public and documenting it :) I just pushed that. Let me know if you find it clear enough

ogrisel: > I just pushed that. Let me know if you find it clear enough

where is this? 479b111 does not explain that the filename of dynamic libraries is used to match the library and that threadpoolctl inspect this filename via a look-up to the C runtime (via `dl_iterate_phdr` and what happens in `_make_controller_from_path`).

jeremiedbb:
> Once the doc is complete, and assuming the CI is fixed (ModuleNotFoundError: No module named 'mesonpy', apparently a new build time dependency for the BLIS CI entries) +1 for merge.

this error is because I tried to set ``--no-build-isolation``. I just reverted it. You'll see the original error. I've not been able to understand what's wrong with this install so far...

ogrisel:
Have you tried to mamba / pip install ninja? It might be a new runtime dependency of numpy when building in editable mode.

threadpoolctl.py
jeremiedbb: I added something

threadpoolctl.py
jeremiedbb: It's doable but it will add some code complexity because we do check both attrs sometimes before the class is instanciated. I think it's ok to keep both mandatory, one can repeat the same name anyway.
We can still decide to make it optional later, but I'd give this a try first if it's fine with you.

tests/_pyMylib/__init__.py
tomMoral, line 1: If this serves as an example, then we probably need to put some comments here

tests/_pyMylib/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+path = Path(__file__).parent / "my_threaded_lib.so"
+ctypes.CDLL(path)
+
+
+class MyThreadedLibController(LibController):
+    user_api = "my_threaded_lib"
+    internal_api = "my_threaded_lib"
+    filename_prefixes = ("my_threaded_lib",)
```
tomMoral, line 14:
```suggestion
    # names for threadpoolctl's context filtering
    user_api = "my_threaded_lib"
    internal_api = "my_threaded_lib"

    # patterns to identify the name of the linked library to load
    # If a library with a matching filename is linked to the python
    # process, it will be available as a dynlib object in the LibController.
    filename_prefixes = ("my_threaded_lib",)
```

jeremiedbb:
I added all the doc I wanted to add. This is ready for another round of reviews :)

tests/_pyMylib/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def get_num_threads(self):
+        return getattr(self.dynlib, "mylib_get_num_threads")()
```
tomMoral, line 17:
```suggestion
    def get_num_threads(self):
        # This function should return the current number of threads,
        # that is reported by `threadpoolctl.threadpool_info`
        return getattr(self.dynlib, "mylib_get_num_threads")()
```

tests/_pyMylib/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def set_num_threads(self, num_threads):
+        getattr(self.dynlib, "mylib_set_num_threads")(num_threads)
```
tomMoral, line 20:
```suggestion
        # This function limits the max number of threads,
        # when it is called in `threadpoolctl.threadpool_limits`.
        getattr(self.dynlib, "mylib_set_num_threads")(num_threads)
```

tests/_pyMylib/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def get_version(self):
+        get_version = getattr(self.dynlib, "mylib_get_version")
```
tomMoral, line 23:
```suggestion
        # This function returns the version of the linked library
        # if It is exposed. It will be returned in `threadpool_info`.
        get_version = getattr(self.dynlib, "mylib_get_version")
```

tests/_pyMylib/__init__.py
ogrisel: @tomMoral I edited your suggestion directly :)

ogrisel (approved):
+1 as well once @tomMoral inline comments suggestion to make the sample plugin more educational are included.

tomMoral (approved):
LGTM, great feature @jeremiedbb

threadpoolctl.py
```diff
@@ -60,56 +67,318 @@ class _dl_phdr_info(ctypes.Structure):
...
+    `get_version`.
+
+    Threadpoolctl loops through all the loaded shared libraries and tries to match
+    the filename of each library with the `filename_prefixes`. If a match is found, a
+    controller is instantiated and a handler to the library is stored in the `dynlib`
+    attribute as a `ctypes.CDLL` object. It can be used to access the necessary symbols
+    of the shared library to implement the above methods.
+
+    The following information will be exposed in the info dictionary:
+      - user_api : user API.
```
ogrisel, line 95:
```suggestion
      - user_api : standardized API (if any).
```

threadpoolctl.py
```diff
@@ -60,56 +67,318 @@ class _dl_phdr_info(ctypes.Structure):
...
+      - internal_api : internal API.
```
ogrisel, line 96:
```suggestion
      - internal_api : implementation-specific API.
```

threadpoolctl.py
```diff
@@ -60,56 +67,318 @@ class _dl_phdr_info(ctypes.Structure):
...
+        - internal_api : str
+            Usually the name of the library or concrete implementation of some
+            specification, e.g. "openblas" is an implementation of the "blas"
+            specification.
+        - filename_prefixes : tuple
+            Possible prefixes of the shared library's filename that allow to
+            identify the library. e.g. "libopenblas" for libopenblas.so.
+
+    and implement the following methods: `get_num_threads`, `set_num_threads` and
+    `get_version`.
```
ogrisel, line 86:
```suggestion
    `get_version`.

    A library controller can optionally specify the `check_symbols` attribute valued
    with a tuple of `str` in cases where the `filename_prefixes` attribute is not
    specific enough to unambiguously identify a dynamically linked native library.
```

threadpoolctl.py
```diff
@@ -60,56 +67,318 @@ class _dl_phdr_info(ctypes.Structure):
...
+
+    Threadpoolctl loops through all the loaded shared libraries and tries to match
+    the filename of each library with the `filename_prefixes`. If a match is found, a
+    controller is instantiated and a handler to the library is stored in the `dynlib`
+    attribute as a `ctypes.CDLL` object. It can be used to access the necessary symbols
+    of the shared library to implement the above methods.
```
ogrisel, line 92:
```suggestion
    of the shared library to implement the above methods assuming the optional
    preconditions defined in the `check_symbols` attribute are met.
```

ogrisel (approved):
A few more improvement. @jeremiedbb feel free to merge if you agree with them.

threadpoolctl.py
jeremiedbb: This is not True yet. We only check the symbols of ``libblas`` on windows to workaround conda renaming. We could try to implement support for that but I'd rather do that in a separate PR.

ogrisel: Let's keep that for a follow-up PR then.

PR #142, opened 2023-07-07
DOC Warn if coexisting libomp / libiomp on MacOS

jeremiedbb:
related to https://github.com/ContinuumIO/anaconda-issues/issues/13221
and https://github.com/scikit-learn-contrib/imbalanced-learn/issues/984

Looks like it's possible to have some crashes due to both libraries coexisting in the same python program on MacOS afterall.

I removed the linux only filter for the warning, but then it means all macos users installing from conda default channel will get this warning. Is it something that we want ?

### DIFF
--- a/multiple_openmp.md
+++ b/multiple_openmp.md
@@ -29,13 +29,13 @@ program**. For instance, on Linux, we never observed any issue between
 `libgomp` and `libiomp`, which is the most common mix (NumPy with MKL + a
 package compiled with GCC, the most widely used C compiler on that platform).
 
-## Incompatibility between Intel OpenMP and LLVM OpenMP under Linux
+## Incompatibility between Intel OpenMP and LLVM OpenMP
 
 The only unrecoverable incompatibility we encountered happens when loading a
 mix of compiled extensions linked with **`libomp` (LLVM/Clang) and `libiomp`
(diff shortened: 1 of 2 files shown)


ogrisel:
Could you please document the change in the warning in the changelog?

CHANGES.md
```diff
@@ -3,6 +3,11 @@
 
 - Dropped support for Python 3.6 and 3.7.
 
+- A warning is raised on MacOS when threadpoolctl finds both Intel OpenMP and LLVM
+  OpenMP runtimes loaded simultaneously by the same Python program. See details and
+  workarounds at https://github.com/joblib/threadpoolctl/blob/master/multiple_openmp.md.
+  https://github.com/joblib/threadpoolctl/pull/142
+
```
ogrisel, line 10:
```suggestion
- A warning is raised on macOS when threadpoolctl finds both Intel OpenMP and LLVM
  OpenMP runtimes loaded simultaneously by the same Python program. See details and
  workarounds at https://github.com/joblib/threadpoolctl/blob/master/multiple_openmp.md.
  https://github.com/joblib/threadpoolctl/pull/142

```

ogrisel:
Merged, thanks @jeremiedbb!

PR #143, opened 2023-07-12
MAINT make CI fail when linting fails

jeremiedbb:
I noticed that we used to run a black check on the CI but even when it was failing, the CI would still be green because it was run in a script that did not return a failing exit code. I changed things a bit to make linting a separate job

### DIFF
--- a/.azure_pipeline.yml
+++ b/.azure_pipeline.yml
@@ -17,8 +17,9 @@ schedules:
 stages:
 - stage:
   jobs:
-  - job: 'run_linter'
-    displayName: Run linter
+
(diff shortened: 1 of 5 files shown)


jeremiedbb:
the ci failed with commit [84a6391](https://github.com/joblib/threadpoolctl/pull/143/commits/84a639198cd7cab12deb44eccb224b99c40ef54d) and no longer fails after running black.

I tried to update the images but it involves more work to also update the versions of the compilers because the current versions are not available anymore from the standard ppas. Let's keep the purpose of this PR minimal.

PR #148, opened 2023-08-08
Fails on FreeBSD: ld-elf.so.1: Can't find module with TLS index 1 -- Fix

### LINKED ISSUE #125: Fails on FreeBSD: ld-elf.so.1: Can't find module with TLS index 1
This code:
```
from threadpoolctl import threadpool_info
from pprint import pprint
pprint(threadpool_info())
```
fails on FreeBSD 13:
```
ld-elf.so.1: Can't find module with TLS index 1
```

ptrajdos:
I have fixed the issue that generated  ** ld-elf.so.1: Can't find module with TLS index 1 ** on FreeBSD.
Tested on FreeBSD 13.2
This closes #125

### DIFF
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -66,6 +66,7 @@ try:
 except AttributeError:
     _RTLD_NOLOAD = ctypes.DEFAULT_MODE
 
+_RTLD_DEFAULT=-2
 
 class LibController(ABC):
     """Abstract base class for the individual library controllers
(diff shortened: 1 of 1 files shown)


threadpoolctl.py
```diff
@@ -818,9 +819,64 @@ def _load_libraries(self):
...
+                self._make_controller_from_path(filepath)
+            return 0
+
+        c_func_signature = ctypes.CFUNCTYPE(
+            ctypes.c_int,  # Return type
+            ctypes.POINTER(_dl_phdr_info),
+            ctypes.c_size_t,
+            ctypes.c_char_p,
+        )
+        c_match_library_callback = c_func_signature(match_library_callback)
```
tomMoral, line 870: Could you factorize this part in a `make_c_match_library_callback`, used in both ` _find_libraries_with_dl_iterate_phdr_freebsd` and ` _find_libraries_with_dl_iterate_phdr`?

threadpoolctl.py
```diff
@@ -818,9 +819,64 @@ def _load_libraries(self):
...
+    def _find_libraries_with_dl_iterate_phdr_freebsd(self):
+        """Loop through loaded libraries and return binders on supported ones
+
+        This function is expected to work on POSIX system only.
+        This code is adapted from code by Intel developer @anton-malakhov
+        available at https://github.com/IntelPython/smp
+
+        Copyright (c) 2017, Intel Corporation published under the BSD 3-Clause
+        license
+        """
```
tomMoral, line 836: Could you update the docstring to explain why this is different compared to the next function?
Also, would this function work on other POSIX platforms?

tomMoral (commented):
hello,

thanks for looking into this. It looks reasonable but if we want to support FreeBSD, we should have some tests to make sure this is working and we can maintain it.
It seems that there is a [github action to use a VM and make tests run on FreeBSD](https://github.com/marketplace/actions/freebsd-vm), this could be a nice way to test that the current code works (I don't have access to a FreeBSD machine myself).

Most of the rest is mostly formatting: I put a few comments and the linter is failing.

threadpoolctl.py
ptrajdos: After issuing this pull request, I've got response in https://github.com/python/cpython/issues/92828.
There is also information here: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=272992
The issue will be fixed in FreeBSD itself. I think there is no longer a need to insert platform-dependent code in threadpoolctl. 

However, it will be worth checking how it works on other POSIX systems. 
I'll try to check on NetBSD and OpenBSD. It may take a while, since I need to setup some VMs.

ptrajdos: I've checked threadpoolctl on NetBSD -- The bug is not present.
I've also checked calling (C language) **dl_iterate_phdr** from libc under Solaris 11. It also works as expected.
It seems that the issue is related to FreeBSD and its implementation of libc. I think it would be better to wait until FreeBSD is fixed. 
What do you think?

ogrisel: I agree. Thank you very much for your efforts at reproducing the problem on various platform and working with getting the problem fixed upstream. This is really helpful.

outpaddling: This issue is preventing MACS3 from functioning on FreeBSD 13.2, but not on 14.0.  This patch resolves the issue.  Do you think this patch is mature enough to add to the devel/py-threadpoolctl port as a stop-gap until the issue is corrected in 13.x?  If so, I'll do more testing and open a FreeBSD ports PR.

ogrisel: I did not review the details but if all the tests pass on FreeBSD 13.2 with this patch backported to the devel/py-threadpoolctl port, then I think it's fine.

outpaddling: Thanks for the suggestion.  Running a standard pytest command on 13.2, it hangs on the command_line_flag test, consuming 100% CPU.

```
tests/test_threadpoolctl.py::test_libomp_libiomp_warning SKIPPED (Re...) [ 88%]
tests/test_threadpoolctl.py::test_command_line_empty PASSED              [ 90%]
tests/test_threadpoolctl.py::test_command_line_command_flag ^C
```
On 14.0, it does not.  Many tests are skipped on both platforms.  I'm guessing the port is lacking optional dependencies, or something of that nature.  I'll just mark py-macs3 as broken on 13.2.  13.3 is scheduled for release shortly anyway, and the problem does not seem to occur on the latest 13.x snapshot.

outpaddling: I could, but I think it's agreed that this patch is not needed and won't be imported.  It is not necessary on FreeBSD 14 or on the latest snapshot of FreeBSD 13.  Among supported releases, it's only needed on FreeBSD 13.2, which is approaching EOL anyway.  The release cycle for 13.3 begins tomorrow:

[https://www.freebsd.org/releases/13.3R/schedule/](https://www.freebsd.org/releases/13.3R/schedule/)

Thanks for the informative suggestion, though.  It's good to know about faulthandler.

ogrisel: Ok so let's close this PR then.

ogrisel:
As discussed above, we agreed this would be a won't fix upstream in threadpoolctl itself.

PR #153, opened 2023-10-04
CI Fix BLIS jobs

jeremiedbb:
- pin numpy < 2 for now for jobs using BLIS

- ~~install python and other packages at once because not all packages are available for py312 yet.~~
  EDIT: actually not simple because for some jobs we need an env with python installed and then install the other packages with pip. The alternative would be to pin python 3.11 instead of `'*'` and bump to 3.12 when scipy is ready but I'd rather keep it that way and wait for the issue to resolve itself.

### DIFF
--- a/continuous_integration/install_with_blis.sh
+++ b/continuous_integration/install_with_blis.sh
@@ -35,6 +35,7 @@ popd
 # build & install numpy
 git clone https://github.com/numpy/numpy.git
 pushd numpy
+git checkout v1.26.0 # pin numpy < 2 for now
 git submodule update --init
 echo "[blis]
 libraries = blis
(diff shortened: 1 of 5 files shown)


continuous_integration/install_with_blis.sh
```diff
@@ -35,7 +35,7 @@ popd
 # build & install numpy
 git clone https://github.com/numpy/numpy.git
 pushd numpy
-git checkout v1.26.0  # pin numpy < 2 for now
+git checkout v1.26.0 # pin numpy < 2 for now
```
ogrisel, line 38:
```suggestion
git checkout v1.26.0  # pin numpy < 2 for now
```

ogrisel:
I guess we can close now.

PR #155, opened 2023-10-16
Python 3.12 version support and make tests tolerate the presence of OpenMP for system Python

### LINKED ISSUE #146: tests/test_threadpoolctl.py::test_command_line_empty regression with Python 3.12
When running the test suite against Python 3.12.0b4, I'm getting the following test failure:

```pytb
_______________________________________________________ test_command_line_empty _______________________________________________________

    def test_command_line_empty():
        output = subprocess.check_output((sys.executable + " -m threadpoolctl").split())
>       assert json.loads(output.decode("utf-8")) == []
E       AssertionError: assert [{'user_api': 'openmp', 'internal_api': 'openmp', 'prefix': 'libgomp', 'filepath': '/usr/lib/gcc/x86_64-pc-linux-gnu/13/libgomp.so.1.0.0', 'version': None, 'num_threads': 12}] == []
E         Left contains one more item: {'filepath': '/usr/lib/gcc/x86_64-pc-linux-gnu/13/libgomp.so.1.0.0', 'internal_api': 'openmp', 'num_threads': 12, 'prefix': 'libgomp', ...}
E         Full diff:
E           [
E         -  ,
E         +  {'filepath': '/usr/lib/gcc/x86_64-pc-linux-gnu/13/libgomp.so.1.0.0',
E         +   'internal_api': 'openmp',
E         +   'num_threads': 12,
E         +   'prefix': 'libgomp',
E         +   'user_api': 'openmp',
E         +   'version': None},
E           ]

tests/test_threadpoolctl.py:535: AssertionError
```

The test passes with Python 3.11.

ogrisel:
Fixes #146.

Actually our CI does not need any update to test against the latest Python version available on conda-forge.

This PR is mostly about updating `pyproject.toml` and renaming `VERSION_PYTHON` to `PYTHON_VERSION` to sound more correct English.

EDIT: also tolerate the presence of OpenMP when Python is not installed via conda.

### DIFF
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -557,14 +557,7 @@ def test_command_line_empty_or_system_openmp():
     if not managed_by_conda:  # pragma: no cover
         # When using a Python interpreter that is not the one from the conda
         # environment, we should ignore any system OpenMP library.
-        results = [
-            library
-            for library in results
-            if not (
-                library["filepath"].startswith("/usr/lib/")
-                and library["user_api"] == "openmp"
-            )
-        ]
+        results = [r for r in results if r["user_api"] != "openmp"]
     assert results == []
 
 
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -545,9 +545,27 @@ def test_libomp_libiomp_warning(recwarn):
     assert "multiple_openmp.md" in str(wm.message)
 
 
-def test_command_line_empty():
+def test_command_line_empty_or_system_openmp():
+    # When the command line is called without arguments, no library should be
+    # detected. The only exception is a system OpenMP library that can be
+    # linked to the Python interpreter, for instance via the libb2.so BLAKE2
+    # library that can itself be linked to an OpenMP runtime on Gentoo.
     output = subprocess.check_output((sys.executable + " -m threadpoolctl").split())
-    assert json.loads(output.decode("utf-8")) == []
+    results = json.loads(output.decode("utf-8"))
+    conda_prefix = os.getenv("CONDA_PREFIX")
+    managed_by_conda = conda_prefix and sys.executable.startswith(conda_prefix)
+    if not managed_by_conda:  # pragma: no cover
+        # When using a Python interpreter that is not the one from the conda
+        # environment, we should ignore any system OpenMP library.
+        results = [
+            library
+            for library in results
+            if not (
+                library["filepath"].startswith("/usr/lib/")
+                and library["user_api"] == "openmp"
+            )
+        ]
+    assert results == []
 
 
 def test_command_line_command_flag():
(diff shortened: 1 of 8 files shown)


tests/test_threadpoolctl.py
```diff
@@ -545,9 +545,27 @@ def test_libomp_libiomp_warning(recwarn):
...
+    conda_prefix = os.getenv("CONDA_PREFIX")
+    managed_by_conda = conda_prefix and sys.executable.startswith(conda_prefix)
+    if not managed_by_conda:  # pragma: no cover
+        # When using a Python interpreter that is not the one from the conda
+        # environment, we should ignore any system OpenMP library.
+        results = [
+            library
+            for library in results
+            if not (
+                library["filepath"].startswith("/usr/lib/")
```
mgorny, line 564: Well, technically this works for us in Gentoo but I think you may need to account for the OMP library residing in `/usr/lib64` or another per-arch libdir as well. Then there's Gentoo Prefix that actually uses `/some-user-defined-prefix/usr/lib`, and NixOS with their horror of paths ;-).

ogrisel: Thanks for the feedback. I simplified the tests to be more tolerant to variations.

tests/test_threadpoolctl.py
```diff
@@ -545,9 +545,20 @@ def test_libomp_libiomp_warning(recwarn):
...
+    # detected. The only exception is a system OpenMP library that can be
+    # linked to the Python interpreter, for instance via the libb2.so BLAKE2
+    # library that can itself be linked to an OpenMP runtime on Gentoo.
     output = subprocess.check_output((sys.executable + " -m threadpoolctl").split())
-    assert json.loads(output.decode("utf-8")) == []
+    results = json.loads(output.decode("utf-8"))
+    conda_prefix = os.getenv("CONDA_PREFIX")
+    managed_by_conda = conda_prefix and sys.executable.startswith(conda_prefix)
+    if not managed_by_conda:  # pragma: no cover
+        # When using a Python interpreter that is not the one from the conda
```
ogrisel, line 558:
```suggestion
        # When using a Python interpreter that does not come from a conda
```

tests/test_threadpoolctl.py
mgorny: Thanks. I can confirm that this version works too.

ogrisel:
@jeremiedbb this PR is good to go. It should speed-up the conda install of other PRs because the default solver is really really slow for some reason.

jeremiedbb (approved):
LGTM. We could also just stop assuming that none of the libraries of interest are linked to python itself and just remove this test, don't you think ?

ogrisel:
> We could also stop assuming that none of the libraries of interest are linked to python itself and just remove this test, don't you think ?

I think we can merge this PR as it is for now and remove this test if this happens gain in another case.

Still, it's interesting to discover that our assumptions about Python runtime dependencies were wrong. This might be important to inform us about choices about packaging such native libraries, see for instance https://github.com/numpy/numpy/issues/24857.

PR #160, opened 2024-01-17
MAINT Update BLIS install to use numpy's new build system

jeremiedbb:


ogrisel (approved):
I checked the output of the `python -m threadpoolctl -i numpy scipy.linalg tests._openmp_test_helper.openmp_helpers_inner` command run on the CI at the beginning of the test step and the output seems to match the intended environment as specified in the YAML config of the CI.

PR #161, opened 2024-01-29
Limit version of flit_core specified to build with

takluyver:
I suggest every package built with Flit specifies a maximum version, so old packages can continue to be built in the future, even if things change in newer versions of Flit.

The recommendations around this are here: https://flit.pypa.io/en/stable/pyproject_toml.html

As a concrete example, the Flit-specific `[tool.flit.metadata]` table is gradually being replaced by the standardised `[project]` table. Flit 3.x has supported both options, but my plan for v4 is to only support `[project]`.


takluyver:
The CI failure is not related to this change - I guess `black` has changed a bit of it's style for 2024.

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["flit_core"]
+requires = ["flit_core >=2,<4"]
 build-backend = "flit_core.buildapi"
 
 [tool.flit.metadata]
```
ogrisel, line 5:
```suggestion
# TODO: replace the following section by the standard [project] section
# to be able to use flit v4.
[tool.flit.metadata]
```

ogrisel (approved):
Thanks for the PR. LGTM with the following TODO:

jeremiedbb (approved):
Thanks @takluyver

PR #163, opened 2024-02-07
FEA Add support to switch FlexiBLAS backend

jeremiedbb:


threadpoolctl.py
```diff
@@ -105,27 +105,25 @@ class LibController(ABC):
...
         self.dynlib = ctypes.CDLL(filepath, mode=_RTLD_NOLOAD)
         self.version = self.get_version()
         self.set_additional_attributes()
 
-    @final
     def info(self):
-        """Return relevant info wrapped in a dict
-
-        This is not meant to be overriden by subclasses.
-        """
```
jeremiedbb, line 121: we now need to be able to override this since the flexiblas controller has dynamic properties (so we can't rely on `vars`)

threadpoolctl.py
```diff
@@ -383,6 +397,36 @@ def _get_current_backend(self):
...
+                load_func = getattr(
+                    self.dynlib, "flexiblas_load_backend_library", lambda _: -1
+                )
+            res = load_func(str(backend).encode("utf-8"))
+            if res == -1:
+                raise RuntimeError(f"Failed to load backend {backend!r}.")
+
+            # Trigger a new search of loaded shared libraries since loading a new
+            # backend caused a dlopen.
+            self.parent._load_libraries()
```
jeremiedbb, line 422: adding support to switch backends comes with a cost. When loading a new backend, we need to rerun the dl_iterate_phdr search for shared libraries.

We specifically designed the ThreadpoolController to avoid rerunning the search every time we call `info` or `limit`, for the reason that "it's unlikely that the loaded shared libraries will change during the lifetime of the program". This is no longer true with FlexiBLAS.

I think the most reasonable thing to do is to only trigger a new search when loading a new backend, even though it makes the LibControllers have a ref to the parent ThreadpoolController (which is the one able to rerun the search).

jeremiedbb (commented):
This feature comes with a cost since it extends the functionalities of threadpoolctl beyond its original design, as illustrated below

jeremiedbb:
question is: do we want do document it or keep it kind of hidden for now ?

threadpoolctl.py
ogrisel: > adding support to switch backends comes with a cost. When loading a new backend, we need to rerun the dl_iterate_phdr search for shared libraries.

I think this is fine as loading a new backend should not happen very often, or at least not in performance critical sections of a program.

> I think the most reasonable thing to do is to only trigger a new search when loading a new backend, even though it makes the LibControllers have a ref to the parent ThreadpoolController (which is the one able to rerun the search).

I agree, although I have not yet reviewed the diff yet.

ogrisel:
> question is: do we want do document it or keep it kind of hidden for now ?

I think we should document it, probably in the readme. Maybe we can state explicitly that this part of the API is considered experimental/unstable for 1 year after the next release.

threadpoolctl.py
```diff
@@ -927,6 +971,7 @@ def __len__(self):
 
     def _load_libraries(self):
         """Loop through loaded shared libraries and store the supported ones"""
+        self.lib_controllers = []  # reset the list of controllers
```
ogrisel, line 974: This will be a bit weird in the case of flexiblas: when we switch backend, the controller instance of flexiblas itself will no longer be the controller used by the caller to trigger the switch.

Maybe this is not that a big problem but it might be a bit surprising and maybe cause weird bugs.

The alternative would be to not reset the previously loaded controllers and change `_make_controller_from_path` to only append to the list of controllers controllers that match previously unseen filepaths.

jeremiedbb: Good point !

> The alternative would be to not reset the previously loaded controllers and change _make_controller_from_path to only append to the list of controllers controllers that match previously unseen filepaths.

I did that

continuous_integration/install_with_flexiblas.sh
```diff
@@ -10,7 +10,7 @@ popd
 conda update -n base conda conda-libmamba-solver -q --yes
 conda config --set solver libmamba
 conda create -n $VIRTUALENV -q --yes -c conda-forge python=$PYTHON_VERSION \
-    pip cython openblas meson-python pkg-config compilers
+    pip cython openblas mkl meson-python pkg-config compilers
```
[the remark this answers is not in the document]
jeremiedbb: Ok I added this var in the azure pipelines config

threadpoolctl.py
```diff
@@ -383,6 +397,36 @@ def _get_current_backend(self):
...
+        if backend not in self.loaded_backends:
+            if backend in self.available_backends:
+                load_func = getattr(self.dynlib, "flexiblas_load_backend", lambda _: -1)
+            else:  # assume backend is a path to a shared library
+                load_func = getattr(
+                    self.dynlib, "flexiblas_load_backend_library", lambda _: -1
+                )
+            res = load_func(str(backend).encode("utf-8"))
+            if res == -1:
+                raise RuntimeError(f"Failed to load backend {backend!r}.")
```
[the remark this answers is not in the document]
jeremiedbb: I added details in the error message but it's hard to specialize the error too much because the user gives a string and we don't know if it's meant to be a name or a path. I don't think it's worth adding too much complexity to try to infer that.

threadpoolctl.py
```diff
@@ -105,27 +105,25 @@ class LibController(ABC):
     """
 
     @final
-    def __init__(self, *, filepath=None, prefix=None):
+    def __init__(self, *, filepath=None, prefix=None, parent=None):
         """This is not meant to be overriden by subclasses."""
+        self.parent = parent
```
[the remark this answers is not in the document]
jeremiedbb: An alternative would be to make it a method of ThreadpoolController but I don't like it because it's a flexiblas specific feature

README.md
```diff
@@ -192,6 +192,89 @@ decorators are accessible through their `wrap` method:
...
+  'architecture': 'Haswell'},
+ {'user_api': 'blas',
+  'internal_api': 'mkl',
+  'num_threads': 2,
+  'prefix': 'libmkl_rt',
+  'filepath': '/home/jeremie/miniforge/envs/flexiblas_threadpoolctl/lib/libmkl_rt.so.2',
+  'version': '2024.0-Product',
+  'threading_layer': 'gnu'}]
+```
+
```
ogrisel, line 277:
```suggestion

You can observe that the previously linked OpenBLAS shared object stays linked to
the Python program indefinitely, but flexiblas itself no longer delegate BLAS calls
to OpenBLAS by looking at the value of the `current_backend` attribute.
```

ogrisel (approved):
LGTM once the suggestion above is considered.

Issue #162, opened 2024-02-02
Tests fails on Mac ARM and packages from macports


nilason:
Running the tests in a MacPorts setting, on an Apple ARM machine, the architecture test fails:

```
Executing:  cd "/opt/local/var/macports/build/_Users_nilason_ports_python_py-threadpoolctl/py311-threadpoolctl/work/threadpoolctl-3.2.0" && /opt/local/Library/Frameworks/Python.framework/Versions/3.11/bin/python3.11 -m installer /opt/local/var/macports/build/_Users_nilason_ports_python_py-threadpoolctl/py311-threadpoolctl/work/threadpoolctl-3.2.0-py3-none-any.whl --destdir /opt/local/var/macports/build/_Users_nilason_ports_python_py-threadpoolctl/py311-threadpoolctl/work/destroot 
--->  Testing py311-threadpoolctl
Executing:  cd "/opt/local/var/macports/build/_Users_nilason_ports_python_py-threadpoolctl/py311-threadpoolctl/work/threadpoolctl-3.2.0" && py.test-3.11 -o addopts='' 
========================================================================= test session starts ==========================================================================
platform darwin -- Python 3.11.7, pytest-7.4.3, pluggy-1.3.0
rootdir: /opt/local/var/macports/build/_Users_nilason_ports_python_py-threadpoolctl/py311-threadpoolctl/work/threadpoolctl-3.2.0
plugins: env-1.1.3, cov-4.1.0, xdist-3.3.1, anyio-4.2.0
collected 63 items

tests/test_threadpoolctl.py ....s.sss.ssssssss.ssssss..s..s.......sssssss.sssss..sss...F..s                                                                      [100%]

=============================================================================== FAILURES ===============================================================================
__________________________________________________________________________ test_architecture ___________________________________________________________________________

    def test_architecture():
        expected_openblas_architectures = (
            # XXX: add more as needed by CI or developer laptops
            "armv8",
            "Haswell",
            "Prescott",  # see: https://github.com/xianyi/OpenBLAS/pull/3485
            "SkylakeX",
            "Sandybridge",
            "VORTEX",
            "Zen",
        )
        expected_blis_architectures = (
            # XXX: add more as needed by CI or developer laptops
            "skx",
            "haswell",
        )
        for lib_info in threadpool_info():
            if lib_info["internal_api"] == "openblas":
>               assert lib_info["architecture"] in expected_openblas_architectures
[...]

Issue #135, opened 2023-03-25
Detect the Apple Accelerate runtime


ogrisel:
It is now possible to install numpy, scipy... with a BLAS implementation that links to the macOS Accelerate library via conda-forge:

- https://github.com/conda-forge/numpy-feedstock/issues/253

But at the moment, threadpoolctl does not recognize this BLAS implementation. I tested the following on a macOS/M1 machines:

```
mamba create -n accelerate numpy "libblas=*=*accelerate" threadpoolctl
mamba activate accelerate
python -m threadpoolctl -i numpy
[]
```

Output of `mamba list`:

<details>

```
# packages in environment at /Users/ogrisel/mambaforge/envs/accelerate:
#
# Name                    Version                   Build  Channel
bzip2                     1.0.8                h3422bc3_4    conda-forge
ca-certificates           2022.12.7            h4653dfc_0    conda-forge
libblas                   3.9.0           16_osxarm64_accelerate    conda-forge
libcblas                  3.9.0           16_osxarm64_accelerate    conda-forge
libcxx                    15.0.7               h75e25f2_0    conda-forge
libffi                    3.4.2                h3422bc3_5    conda-forge
libgfortran               5.0.0           12_2_0_hd922786_31    conda-forge
libgfortran5              12.2.0              h0eea778_31    conda-forge
liblapack                 3.9.0           16_osxarm64_accelerate    conda-forge
libsqlite                 3.40.0               h76d750c_0    conda-forge
libzlib                   1.2.13               h03a7124_4    conda-forge
llvm-openmp               16.0.0               h7cfbb63_0    conda-forge
ncurses                   6.3                  h07bb92c_1    conda-forge
numpy                     1.24.2          py311h60f8152_0    conda-forge
openssl                   3.1.0                h03a7124_0    conda-forge
pip                       23.0.1             pyhd8ed1ab_0    conda-forge
python                    3.11.0          h3ba56d0_1_cpython    conda-forge
python_abi                3.11                    3_cp311    conda-forge
readline                  8.2                  h92ec313_1    conda-forge
setuptools                67.6.0             pyhd8ed1ab_0    conda-forge
threadpoolctl             3.1.0              pyh8a188c0_0    conda-forge
tk                        8.6.12               he1e0b03_0    conda-forge
tzdata                    2023b                h71feb2d_0    conda-forge
wheel                     0.40.0             pyhd8ed1ab_0    conda-forge
[...]

PR #166, opened 2024-02-09
[WIP] FEA Support Veclib/Accelerate

jeremiedbb:


jeremiedbb:
I don't think it's possible to support Accelerate:
- when installed by conda-forge, there's only a libblas.dylib (whereas openblas for instance comes with libblas.dylib and libopenblas.dylib) so threadpoolctl can't rely on the lib name only to guess which one it is.
- this libblas.dylib seems to not expose any symbol that allows to identify accelerate. It only exposes the generic blas symbols. I say that based on running `nm -a libblas.dylib`, where the only non generic symbol `_veclib` which is a local (non-external) bss (block starting symbol) section symbol which is not accessible in python using hasattr.

ogrisel:
When using otool I can see that it's actually linked to `libvecLibFort-ng` via an @rpath entry:

```
otool -L /Users/ogrisel/mambaforge-disabled/envs/dev/lib/libblas.3.dylib
/Users/ogrisel/mambaforge-disabled/envs/dev/lib/libblas.3.dylib:
        @rpath/libvecLibFort-ng.dylib (compatibility version 0.0.0, current version 0.0.0)
        /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib (compatibility version 1.0.0, current version 1.0.0, reexport)
        @rpath/liblapack-netlib.3.9.0.dylib (compatibility version 0.0.0, current version 0.0.0, reexport)
        @rpath/liblapacke-netlib.3.9.0.dylib (compatibility version 0.0.0, current version 0.0.0, reexport)
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1292.0.0)
```

ogrisel:
Here are known cases of numpy being linked accelerate in the wild:

- numpy installed from conda-forge with `conda install -c conda-forge numpy libblas=*=*accelerate"`
- the official numpy wheels for macOS on pypi.org as discussed in https://github.com/joblib/threadpoolctl/issues/135#issuecomment-1782572823 (not the case at the time of writing though, see the linked discussion);
- the macports numpy package: https://github.com/joblib/threadpoolctl/issues/162#issuecomment-1932138190

Note that the various packaging methods my differ in how vecLib is linked to the Python program. This might introduce additional difficulty in finding a robust heuristic to detect accelerate/vecLib given the fact that there are no vecLib-specific exported symbols, as far as I know.

jeremiedbb:
So irl we discussed the possibility to check if the detected `libblas.dylib` has a dependency over `libvecLibFort-ng.dylib`, but I didn't manage to do that easily.

Although on linux it's seems doable using `dlinfo`, this is gnu only, and I didn't an equivalent for macos. The `dyld` functions seems too limited (https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/dyld.3.html). 

I tried to look at the source code of `otool` to find how they do it but the solution seems way to complicated...

I have an alternative that I don't particularly like. When we find a `libblas` we create a subprocess where we only dlopen this `libblas` and check if `_find_libraries_with_dyld` finds `libvecLibFort-ng.dylib`. It would not add that much complexity to the code base but the cost of creating a new process is clearly not negligible.

jeremiedbb:
Another alternative that we discussed irl but I'm not a fan either is to directly check for `libvecLibFort-ng.dylib` but it's not the real dylib containing the blas, and the info dict would not show the path to libblas either.

jeremiedbb:
Anyway, let's not delay the release any longer for that.

ogrisel:
conda-forge recently introduced: https://conda-forge.org/news/2025/07/31/new-accelerate-macos/

Not sure if it changes anything w.r.t. discoverability.

PR #169, opened 2024-02-15
Add support for Pyodide

### LINKED ISSUE #167: Remove libc warning inside Pyodide?
Inside Pyodide [0.25 console](https://pyodide.org/en/stable/console.html):
```py
import threadpoolctl
threadpoolctl.threadpool_info()
```

```
/lib/python3.11/site-packages/threadpoolctl.py:1019: RuntimeWarning: libc not found. The ctypes module in Python 3.11 is maybe too old for this OS.
```

This can cause a confusing warning when running scikit-learn code so maybe we should consider not warning in this case?

image

`sys.platform == 'emscripten'` seems like a simple way to check that we are inside Pyodide (more generally Python has been compiled for emscripten). See https://pyodide.org/en/stable/usage/faq.html#how-to-detect-that-code-is-run-with-pyodide for more details.

lesteve:
I have done a quick an dirty test inside Pyodide that seems to indicate that it works but I want to double-check rebuilding Pyodide up-to scikit-learn with my threadpoolctl branch to make sure that `sklearn.show_versions()` works

Not sure if this is worth testing in the CI ...

Closes #167

### DIFF
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -979,6 +979,8 @@ class ThreadpoolController:
             self._find_libraries_with_dyld()
         elif sys.platform == "win32":
             self._find_libraries_with_enum_process_module_ex()
+        elif sys.platform == "emscripten":
+            self._find_libraries_emscripten()
         else:
             self._find_libraries_with_dl_iterate_phdr()
 
@@ -1100,6 +1102,27 @@ class ThreadpoolController:
         finally:
             kernel_32.CloseHandle(h_process)
 
+    def _find_libraries_emscripten(self):
+        """Emscripten specific implementation taken from Pyodide CPython patch for ctypes.util.find_library
+        """
+        libraries = []
+        ld_library_paths = [
+            path
+            for path in os.environ.get("LD_LIBRARY_PATH", "").split(":")
+            if os.path.exists(path)
+        ]
+        for dir in ld_library_paths:
+            dir_full_paths = [os.path.join(dir, name) for name in os.listdir(dir)]
+            dir_libraries = [
+                path
+                for path in dir_full_paths
+                if path.endswith(".so") or path.endswith(".wasm") and _is_wasm(path)
+            ]
+            libraries += dir_libraries
+
+        for library in libraries:
+            self._make_controller_from_path(library)
+
     def _make_controller_from_path(self, filepath):
         """Store a library controller if it is supported and selected"""
         # Required to resolve symlinks
@@ -1192,12 +1215,13 @@ class ThreadpoolController:
         if libc is None:
             libc_name = find_library("c")
             if libc_name is None:  # pragma: no cover
-                warnings.warn(
-                    "libc not found. The ctypes module in Python"
-                    f" {sys.version_info.major}.{sys.version_info.minor} is maybe"
-                    " too old for this OS.",
-                    RuntimeWarning,
-                )
+                if sys.platform != "emscripten":
+                    warnings.warn(
+                        "libc not found. The ctypes module in Python"
+                        f" {sys.version_info.major}.{sys.version_info.minor} is maybe"
+                        " too old for this OS.",
+                        RuntimeWarning,
+                    )
                 return None
             libc = ctypes.CDLL(libc_name, mode=_RTLD_NOLOAD)
             cls._system_libraries["libc"] = libc
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -1118,6 +1118,7 @@ class ThreadpoolController:
             warnings.warn(
                 "Unable to import LDSO from pyodide_js._module. This should never happen"
             )
+            return
 
         for filepath in LDSO.loadedLibsByName.as_object_map():
             # Some libraries are duplicated by Pyodide and do not exist in the
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -1119,7 +1119,11 @@ class ThreadpoolController:
                 "Unable to import LDSO from pyodide_js._module. This should never happen"
             )
 
-        for filepath in LDSO.loadedLibsByName.as_object_map().keys():
+        for filepath in LDSO.loadedLibsByName.as_object_map():
+            # Some libraries are duplicated by Pyodide and do not exist in the
+            # filesystem, so we first check for the existence of the file. For
+            # more details, see
+            # https://github.com/joblib/threadpoolctl/pull/169#issuecomment-1947800866
             if os.path.exists(filepath):
                 self._make_controller_from_path(filepath)
 
--- a/threadpoolctl.py
+++ b/threadpoolctl.py
@@ -1109,9 +1109,11 @@ class ThreadpoolController:
         """
         try:
             from pyodide_js._module import LDSO
-            return LDSO.loadedLibsByName.as_object_map().keys()
         except ImportError:
-            return []
+            warnings.warn("Unable to import LDSO from pyodide_js._module. This should never happen")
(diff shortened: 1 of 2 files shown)


lesteve:
This seems to work inside a Pyodide environment with my threadpoolctl branch:
image

The difference with the current behaviour inside Pyodide:

image

threadpoolctl.py
```diff
@@ -1100,6 +1102,37 @@ def _find_libraries_with_enum_process_module_ex(self):
...
+            dir_full_paths = [os.path.join(dir, name) for name in os.listdir(dir)]
+            dir_libraries = [
+                path
+                for path in dir_full_paths
+                if path.endswith(".so") or path.endswith(".wasm") and _is_wasm(path)
+            ]
+            libraries += dir_libraries
+
+        for library in libraries:
+            self._make_controller_from_path(library)
```
jeremiedbb, line 1134: Could you paste the output of this ? I'd be interested in seing all the libs it finds this way
```suggestion
        for library in libraries:
            print(library)
            self._make_controller_from_path(library)
```

jeremiedbb (commented):
looks promising, but it's a bit different of how we look for libs on other platforms and I think this way may find false positives since it does not search into the actually loaded libs.

lesteve:
> I think this way may find false positives since it does not search into the actually loaded libs.

Not sure how library loading works behind the scenes but it does look like it only finds loaded libraries. A stand-alone snippet that can be copy-and-pasted in [Pyodide console](https://pyodide.org/en/latest/console.html):

```py
import os


def _is_wasm(filename):
    # Return True if the given file is an WASM module
    wasm_header = b"\x00asm"
    with open(filename, "br") as thefile:
        return thefile.read(4) == wasm_header


def list_libraries():
    libraries = []
    ld_library_paths = [
        path
        for path in os.environ.get("LD_LIBRARY_PATH", "").split(":")
        if os.path.exists(path)
    ]
    for dir in ld_library_paths:
        dir_full_paths = [os.path.join(dir, name) for name in os.listdir(dir)]
        dir_libraries = [
            path
            for path in dir_full_paths
            if path.endswith(".so") or path.endswith(".wasm") and _is_wasm(path)
        ]
        libraries += dir_libraries
    return libraries

print('before import scipy', list_libraries())
import scipy
print('after import scipy', list_libraries())
```

Output in Pyodide console:
```
before import scipy []
after import scipy ['/usr/lib/libopenblas.so']
```

lesteve:
I have asked the Pyodide people some questions about my approach in https://github.com/pyodide/pyodide/discussions/4524.

Feel free to jump in there if you have additional questions/concerns!

jeremiedbb:
Hum so actually it misses a lot of libs this way. To tell the dynamic loader where to find the libraries there are different possibilities. One is ``LD_LIBRARY_PATH`` which contains a list of directories to search. Another one is to set the ``rpath`` at compile time because the lib is in a non conventional place and you don't want to pollute ``LD_LIBRARY_PATH`` for instance. 

We could assume that the BLAS libs will always be found by ``LD_LIBRARY_PATH`` but I'm sure how robust this is.

hoodmane:
I'd recommend using the following as long as `dl_iterate_phdr` doesn't work:
```py
def loaded_libs_names_iterator():
   from pyodide_js._module import LDSO
   return LDSO.loadedLibsByName.as_object_map().keys()
```

lesteve:
@hoodmane it seems like there are multiple libopenblas.so in the `LDSO` object: one key is '/usr/lib/libopenblas.so` the other is 'libopenblas.so`, not sure whether that is expected. For now I am checking that the file exists to avoid an error.

```py
import scipy
from pyodide_js._module import LDSO
loaded_libraries = list(LDSO.loadedLibsByName.as_object_map().keys())
print([x for x in loaded_libraries if 'openblas' in x])
```

```
['/usr/lib/libopenblas.so', 'libopenblas.so']
```

hoodmane:
Oh yeah I think we double up the entries for libraries as a hack to work around some library loading bugs, but I can't remember why exactly. I think you should skip the entries that don't start with a `/`.

ryanking13:
> LDSO.loadedLibsByName

yes, I also think `LDSO.loadedLibsByName` is more reliable than `ctype.util.find_library` if what you want is to find loaded libs, as it is possible that shared libs are loaded from other places (e.g. inside the wheel (.libs)).

> one key is /usr/lib/libopenblas.so the other is libopenblas.so, not sure whether that is expected. 

Yes, this is intended. The reason for this is to prevent duplicate loading of already loaded libraries.

For example, say that openblas.so has been loaded from `/usr/lib/openblas.so`, then loadedLibsByName  object will store `/usr/lib/openblas.so`. Then, say that we have another library e.g. scipy.so that depends on openblas.so, and you load scipy.so, emscripten will look for `openblas.so` in loadedLibsByName, not `/usr/lib/openblas.so`. Therefore, we manually add `openblas.so` into the loadedLibsByName.

Related code pointer: https://github.com/pyodide/pyodide/blob/909b22497adb9597e0aac8b37c089fe2cbe0a965/src/js/dynload.ts#L118-L129

We can probably think of ways to improve this, but for now, this was the way for me to make things work.

lesteve:
Thanks for the details @ryanking13! I think this is perfectly fine to adapt to this on the threadpoolctl side.

jeremiedbb:
Looks good, thanks !
I wonder if we should add a Pyodide CI job, especially since we're using some kind of private stuff that is subject to change.

lesteve:
> I wonder if we should add a Pyodide CI job, especially since we're using some kind of private stuff that is subject to change.

This may be doable but this needs a bit of work and I am not sure this is worth it right now.

Probably the simplest thing to do would be to look at numpy CI which is the simplest (`python` and `pip` are wrapper script that makes you believe you are in a normal venv but actually execute nodejs code behind the scenes to run Python code inside Pyodide)

jeremiedbb:
yeah maybe it can wait. Can you try this though to make sure it doesn't crash:
```py
import numpy as np
from threadpoolctl import threadpool_info, threadpool_limits

a = np.random.random_sample((100, 100))
a @ a

print(threadpool_info())

with threadpool_limits(limits=2):
    print(threadpool_info())
    a @ a
```

lesteve:
So right now, in Pyodide numpy does not use OpenBLAS see https://github.com/pyodide/pyodide/issues/3763. What would be an equivalent snippet with scipy?

jeremiedbb:
Ah I see, I updated it, we actually don't need to run any computation

jeremiedbb (approved):
LGTM

threadpoolctl.py
```diff
@@ -1100,6 +1102,32 @@ def _find_libraries_with_enum_process_module_ex(self):
...
+        One day, we may have a simpler solution. libc dl_iterate_phdr needs to
+        be implemented in Emscripten and exposed in Pyodide, see
+        https://github.com/emscripten-core/emscripten/issues/21354 for more
+        details.
+        """
+        try:
+            from pyodide_js._module import LDSO
+        except ImportError:
+            warnings.warn(
+                "Unable to import LDSO from pyodide_js._module. This should never happen"
```
ogrisel, line 1119:
```suggestion
                "Unable to import LDSO from pyodide_js._module. This should never "
                "happen."
```

ogrisel (approved):

hoodmane (approved):
Thanks everyone! This looks good from my perspective. I will work on implementing `dl_iterate_phdr` in emscripten and hopefully in another year or so we can drop this.

PR #184, opened 2025-03-06
CI Add gh workflow for linting

jeremiedbb:
Starting CI migration from azure pipelines to gh actions


jeremiedbb:
Merging since it will not run in the PR as a first workflow anyway.

PR #199, opened 2025-06-11
Static typing

jorenham:
I noticed that `threadpoolctl` did not have static typing support, so I quicky wrote some (non-intrusive) stubs. The `.pyi` stub covers 100% of the public api, and (almost?) everything of the private API.

I went with stubs instead of inline annotations because I noticed that performance matters, and inline annotations tend to negatively impact that. Another reason is that I'm (very) used to writing stubs, so I'm pretty biased haha. Note that there some automated tools available that merge `.pyi` stubs into a `.py` as inline annotations.

I must admit that I'm not very familiar with the codebase, so there's a chance that I made a small mistake here and there.  So please let me know if you see anything questionable. AFAIK it's also fine to commit to this branch yourself.

Mypy's `stubtest` tool (the state-of-the art for validating stubs) seems to like it. I added a `stubtest` job to the CI which will help prevent any regressions.

I wasn't able to get `flit` to include `threadpoolctl.pyi` in the wheel with the original "flat" project layout. So I unfortunately had to change it to a package-based layout. The sdist and wheel are building juist fine, and include the `.pyi`  and the `py.typed`. The need for a   `py.typed` is explained at  https://typing.python.org/en/latest/spec/distributing.html#packaging-typed-libraries . Note that even if the `.pyi` annotations were to be inlined in the `.py`, then you'd still need the `py.typed`, because otherwise static type-checkers will simply ignore the annotations. And I don't think that it's possible to package a `py.typed` if you have a "flat" project layout, but I might be wrong.


jorenham:
ready for review :)

jorenham:
I moved the CLI logic into a new `__main__.py` module. Locally `python -m threadpoolctl` seems to be working again.

jorenham:
> What is the expected performance difference between the 2 options.

Somewhere between 50 and 200 ms startup time, mostly because of the additional imports.

PR #200, opened 2025-09-14
Static typing - inline annotations

jorenham:
This is an alternative approach to #199.

I managed to avoid importing `collections.abc`, `types`, and `typing_extensions` by using stringified annotations. There's no way to get around the `typing` import though, so a `import threadpoolctl` could still be slightly slower for projects that didn't already import `typing`.

Personally, I like this better than #199, and it's not as messy as I had expected it might be. 

Compared to #199, this is also compatible with more type-checkers, namely pyright, basedpyright (a stricter pyright fork), and ty (by the same authors as ruff and uv). Mypy and the pyrights are configured in strict mode, meaning that they'll complain if they aren't able to infer some type. So that's 100% type-coverage.

I tried my best not to touch the the runtime code for as much as I could. In some cases I had to make slight adjustments, but those should not have any effect on the behavior.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -272,3 +272,22 @@ jobs:
       - name: Check no test always skipped
         run: |
           python continuous_integration/check_no_test_skipped.py test_results
+
+  typecheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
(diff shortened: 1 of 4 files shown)


johnklehm-attain:
@ogrisel thoughts here?

PR #203, opened 2025-11-09
Fix main CI

lesteve:
### Windows error

So using `llvm-openmp=20.1.8=*_1` fixes the Windows issue with having both libomp and libiomp.

See https://github.com/conda-forge/openmp-feedstock/pull/184/files with this comment

```
:: this package creates a copy of 'libomp.dll' and installs it as 'libiomp5md.dll' to
:: replace Intel's openmp library. The issue with copying is that if a process loads
:: both 'libimp5md.dll' and 'libomp.dll', one of them will complain that another OpenMP
:: runtime is being loaded even though they are literally the same.
:: To avoid this, let's make 'libiomp5md.dll' a DLL that forwards to 'libomp.dll'
```

Not sure what threadpoolctl should do here ...

### macOS FlexiBLAS issue

TBC


ogrisel:
Closing this in favor of new PRs that fix the same issues.

PR #213, opened 2026-06-23
Thread limit introspection API, part 1: API scope

### LINKED ISSUE #211: API to introspect the scope of what thread pool limit set/get API actually does
There are two things to introspect:

1. Does setting the thread pool limit affect only the current thread, or the whole process?
2. If it affects the whole process, is it setting a per-thread limit for the whole process (what MKL does) or is changing the size of a process-wide global thread pool (OpenBLAS with pthreads)? **UPDATE:** This part is now in #214.

Use cases for introspection:

* Ensuring documentation on individual controllers is correct.
* A debugging tool for users, both of their current setup, and of potential bugs in threadpoolctl.
* Maybe: Getting a sense of what the current FlexiBLAS backend is actually doing when you tell it to set a limit. Given different OpenBLAS behaviors depending on _its_ threading backend, and given limits of FlexiBLAS introspection, this may need to be done at runtime...

itamarst:
Fixes #211.

Add info to determine the scope of the thread-limit-setting API (thread local, process-wide, unknown). Introspecting the semantics of the thread pool are going to be a separate PR, assuming that can be done reasonably.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,10 @@
 3.7.0 (TBD)
 ===========
 
-- Added the ability to check whether a limiting API
-  affects just the current thread or the whole process. Mainly aimed at debugging and diagnostics, and somewhat unreliable, it is therefore enabled by default only for command-line usage.
+- Added the ability to check whether a limiting API affects just the current
+  thread or the whole process. Mainly aimed at debugging and diagnostics, and
+  somewhat unreliable, it is therefore enabled by default only for command-line
+  usage.
   https://github.com/joblib/threadpoolctl/pull/213
 
 - Only warn about simultaneous `libomp` and `libiomp` usage on Linux, where the
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,10 @@
+Unreleased
+==============
+
+- Added the ability to check whether a limiting API
+  affects just the current thread or the whole process. Mainly aimed at debugging and diagnostics, and somewhat unreliable, it is therefore enabled by default only for command-line usage.
+  https://github.com/joblib/threadpoolctl/pull/213
+
 3.6.0 (2025-03-13)
 ==================
 
--- a/tests/test_api_introspection.py
+++ b/tests/test_api_introspection.py
@@ -38,7 +38,7 @@ def test_determine_api_scope_thread_local():
     thread-local implementation.
     """
     api = FakeThreadLocalAPI()
-    assert _determine_api_scope(api.get, api.set) == _APIScope.THREAD_LOCAL
+    assert _determine_api_scope(api.get, api.set) == _APIScope.CURRENT_THREAD
 
 
 @pytest.mark.parametrize("default", [1, 17])
@@ -48,4 +48,4 @@ def test_determine_api_scope_processiwde(default: int):
     process-wide implementation.
     """
     api = FakeProcesswideAPI(default)
-    assert _determine_api_scope(api.get, api.set) == _APIScope.PROCESSWIDE
+    assert _determine_api_scope(api.get, api.set) == _APIScope.PROCESS
--- a/tests/test_api_introspection.py
+++ b/tests/test_api_introspection.py
@@ -2,11 +2,19 @@
 Tests for get/set number of threads API introspection.
 """
 
+import sys
 from threading import local as threadlocal
 
 import pytest
 
-from threadpoolctl import _ThreadLimitScope, _determine_thread_limit_scope
+from threadpoolctl import (
+    _ThreadLimitScope,
+    _determine_thread_limit_scope,
+    ThreadpoolController,
+)
+
+# Make sure we have some BLAS libraries loaded:
+from . import utils as _
 
 
 class FakeThreadLocalAPI(threadlocal):
@@ -52,3 +60,38 @@ def test_determine_thread_limit_scope_processwide(default: int) -> None:
     """
     api = FakeProcesswideAPI(default)
     assert _determine_thread_limit_scope(api.get, api.set) == _ThreadLimitScope.PROCESS
+
+
+@pytest.mark.skipif(
+    sys.platform != "linux", reason="We only hardcoded Linux-specific behavior"
+)
+@pytest.mark.parametrize(
+    ["select_filter", "expected_thread_limit_scope"],
+    [
+        (
+            {"internal_api": "openblas", "threading_layer": "pthreads"},
+            _ThreadLimitScope.PROCESS,
+        ),
+        (
+            {"user_api": "openmp"},
+            _ThreadLimitScope.CURRENT_THREAD,
+        ),
+    ],
+)
+def test_api_scope(
+    select_filter: dict[str, str], expected_thread_limit_scope: str
+) -> None:
+    """
+    Check ``_determine_thread_limit_scope()`` against libraries with known
+    properties, to make sure it detects them correctly.  The test is intended
+    to be of the function's behavior, not of the libraries.
+    """
+    controller = ThreadpoolController().select(**select_filter)
+    if not controller.lib_controllers:
+        pytest.skip(f"{select_filter} controller not found")
+
+    for lib in controller.lib_controllers:
+        assert (
+            _determine_thread_limit_scope(lib.get_num_threads, lib.set_num_threads)
+            == expected_thread_limit_scope
+        )
--- a/tests/test_api_introspection.py
+++ b/tests/test_api_introspection.py
@@ -6,7 +6,7 @@ from threading import local as threadlocal
 
 import pytest
 
-from threadpoolctl import _APIScope, _determine_api_scope
+from threadpoolctl import _APIScope, _determine_api_scope, ThreadpoolController
 
 
 class FakeThreadLocalAPI(threadlocal):
@@ -44,8 +44,35 @@ def test_determine_api_scope_thread_local():
 @pytest.mark.parametrize("default", [1, 17])
 def test_determine_api_scope_processiwde(default: int):
     """
-    Check ``determine_api_scope()`` can correctly diagnose a trivial
+    Check ``_determine_api_scope()`` can correctly diagnose a trivial
     process-wide implementation.
     """
     api = FakeProcesswideAPI(default)
     assert _determine_api_scope(api.get, api.set) == _APIScope.PROCESS
+
+
+@pytest.mark.parametrize(
+    ["select_filter", "expected_api_scope"],
+    [
+        (
+            {"internal_api": "openblas", "threading_layer": "pthreads"},
+            _APIScope.PROCESS,
+        ),
+        ({"user_api": "openmp", "prefix": "libgomp"}, _APIScope.CURRENT_THREAD),
+        ({"user_api": "openmp", "prefix": "libomp"}, _APIScope.CURRENT_THREAD),
+    ],
+)
+def test_api_scope(select_filter: dict[str, str], expected_api_scope: str) -> None:
+    """
+    Check ``determine_api_scope()`` against known values, to make sure it
+    detects them correctly.
+    """
+    controller = ThreadpoolController().select(**select_filter)
+    if not controller.lib_controllers:
+        pytest.skip(f"{select_filter} controller not found")
+
+    for lib in controller.lib_controllers:
+        assert (
+            _determine_api_scope(lib.get_num_threads, lib.set_num_threads)
+            == expected_api_scope
+        )
--- a/tests/test_api_introspection.py
+++ b/tests/test_api_introspection.py
@@ -9,7 +9,6 @@ import pytest
 from threadpoolctl import _APIScope, _determine_api_scope
 
 
-
 class FakeThreadLocalAPI(threadlocal):
     """Thread-local num threads setting API."""
 
--- a/tests/test_api_introspection.py
+++ b/tests/test_api_introspection.py
@@ -3,9 +3,11 @@ Tests for get/set number of threads API introspection.
 """
 
 from threading import local as threadlocal
-from threadpoolctl import APIScope, determine_api_scope
 
 import pytest
+
+from threadpoolctl import _APIScope, _determine_api_scope
+
 
 
 class FakeThreadLocalAPI(threadlocal):
@@ -33,11 +35,11 @@ class FakeProcesswideAPI:
 
 def test_determine_api_scope_thread_local():
     """
-    Check ``determine_api_scope()`` can correctly diagnose a trivial
+    Check ``_determine_api_scope()`` can correctly diagnose a trivial
     thread-local implementation.
     """
     api = FakeThreadLocalAPI()
-    assert determine_api_scope(api.get, api.set) == APIScope.THREAD_LOCAL
+    assert _determine_api_scope(api.get, api.set) == _APIScope.THREAD_LOCAL
 
 
 @pytest.mark.parametrize("default", [1, 17])
@@ -47,4 +49,4 @@ def test_determine_api_scope_processiwde(default: int):
     process-wide implementation.
     """
     api = FakeProcesswideAPI(default)
-    assert determine_api_scope(api.get, api.set) == APIScope.PROCESSWIDE
+    assert _determine_api_scope(api.get, api.set) == _APIScope.PROCESSWIDE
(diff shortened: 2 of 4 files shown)


tests/test_threadpoolctl.py
```diff
@@ -792,3 +794,24 @@ def test_custom_controller():
...
 
     assert ThreadpoolController().info() == original_info
+
+
+@pytest.mark.parametrize(
+    ["select_filter", "expected_api_scope"],
+    [
+        ({"internal_api": "openblas"}, "process"),
+        ({"internal_api": "mkl"}, "process"),
+        ({"internal_api": "blis"}, "process"),
```
ogrisel: Shouldn't the value of `expected_api_scope` also depends on the current threading layer of each BLAS runtime?

threadpoolctl.py
```diff
@@ -69,6 +71,89 @@ class _dl_phdr_info(ctypes.Structure):
...
+    # Some plausible constraints we need to keep in mind:
+    #
+    # 1. The API might not allow setting more than the number of (available, or
+    #    physical) cores.
+    # 2. ...
+    try:
+        # Choose a desired number of threads that is different than the current
+        # number, and hopefully achievable under the current configuration:
+        if previous < 2:
+            expected = 2
```
ogrisel, line 115: Can we safely assume that all runtimes allow for growing the number of threads even if externally limited, e.g. by some environment variables?

I tried with libomp and openblas + pthreads and it seems that I can use `threadpoolctl.threadpool_limits(2)` to grow the number of threads beyond what is set by `OMP_NUM_THREADS=1 OPENBLAS_NUM_THREADS=1` but I wonder if this always the case.

threadpoolctl.py
```diff
@@ -69,6 +71,89 @@ class _dl_phdr_info(ctypes.Structure):
...
+
+        # First, getting in the same thread as a set should always give same
+        # number, if it's a number in a reasonable range. A possible exception
+        # is if the number of thread is limited by available CPU, and only one
+        # CPU is available. In that case we can't empirically determine how the
+        # API works. We try to not reach that point here, but you can imagine a
+        # thread pool implementation that is aware of cgroups, in which case a
+        # Docker container limited to one core will pass the safety check at
+        # the start of the function. Perhaps cpu_count() from loky should be
+        # moved into this package...
```
ogrisel: I agree this will be the best solution, but we can wait until the need actually arises.

threadpoolctl.py
```diff
@@ -69,6 +71,89 @@ class _dl_phdr_info(ctypes.Structure):
...
+
+def _determine_api_scope(
+    get_n_threads: Callable[[], int], set_n_threads: Callable[[int], None]
+) -> _APIScope:
+    """
+    Run some experiments to determine the scope of the given get/set API.
+
+    An attempt will be made to restore all settings to their previous state.
+
+    This won't work if you only have one core available.
```
ogrisel: Since this function can temporarily mutate global process state, it is not thread safe and can have a side effect on concurrently running code that use the libraries being dynamically inspected. I think we should document those limitations explicitly.

ogrisel (commented):
I find it a bit disturbing that calling `threadpoolctl.threadpool_info` can temporarily mutate the libraries global state in a non-thread safe way.

The alternative would be to hardcode expected scopes in `LibController` for all known libraries (and return "unknown" otherwise) and only call `_determine_api_scope` in tests or by passing a non-default kwarg value: `threadpool_ctl.threadpool_info(api_scopes="effective")` (and `api_scopes="expected"` would be the default) for instance.

ogrisel:
Another interesting finding by looking at the CI results: BLIS is `current_thread` scoped contrary to other BLAS implementations.

tests/test_threadpoolctl.py
itamarst: Yeah this was just an initial sketch to see what happens when it runs in CI.

itamarst:
I think it's useful to have it in `python -m threadpoolctl` for debugging and user bug reports, so maybe have the side-effect-y introspection gated behind a keyword argument to `info()`? Or it can just be some add-on processing that the command-line version does using lower-level functions.

ogrisel:
I am fine with introducing an extra CLI flag to enable the extra inspection work.

itamarst:
My feeling is:

* As you said it shouldn't be on by default in the Python API because of risk of side-effects.
* It _should_ be on by default in the CLI because it won't actually run any user code, it's just for debugging, so may as well give all the available info.

threadpoolctl.py
itamarst: Will do.

threadpoolctl.py
itamarst: I imagine we cannot assume that. So I guess I'll document that as another potential limitation.

tests/test_threadpoolctl.py
itamarst: Decided to just rip this out.

threadpoolctl.py
```diff
@@ -69,6 +71,96 @@ class _dl_phdr_info(ctypes.Structure):
     _RTLD_NOLOAD = ctypes.DEFAULT_MODE
 
 
+class _APIScope(Enum):
```
ogrisel: Maybe we could rename to `_ThreadLimitScope` to make this more specific.

threadpoolctl.py
```diff
@@ -69,6 +71,96 @@ class _dl_phdr_info(ctypes.Structure):
...
+    This function might not work if you only have one core available.
+
+    This function might not work if you set a limit on a library with an
+    environment variable.
+
+    The function works by changing the number of threads in loaded controllers,
+    which can be a process-wide change. As such, it is not always thread-safe.
+    An attempt will be made to restore all settings to their previous state,
+    but the result may be subtly different, e.g. if "unset" has different
+    semantics than "set to the default returned value".
```
ogrisel, line 102: Thinking about it a bit more, we could make this inspection side effect free by calling this function in an isolated via `subprocess.run` if we really wanted.

The problem would be to make sure that the same native threadpool libraries that are loaded in the current process at the time of the inspection call are also loaded in the subprocess. To do so we could manually call `ctypes.CDLL` but that might be a bit brittle.

threadpoolctl.py
```diff
@@ -116,15 +208,28 @@ def __init__(self, *, filepath=None, prefix=None, parent=None):
...
         hidden_attrs = ("dynlib", "parent", "_symbol_prefix", "_symbol_suffix")
-        return {
+        result = {
             "user_api": self.user_api,
             "internal_api": self.internal_api,
             "num_threads": self.num_threads,
             **{k: v for k, v in vars(self).items() if k not in hidden_attrs},
         }
+        if extra_info:
+            result["api_scope"] = _determine_api_scope(
```
ogrisel: Similarly, we might want to rename this to `thread_limit_scope` because this scope is mostly about the impact of `set_num_threads` right?

threadpoolctl.py
```diff
@@ -563,8 +668,18 @@ def threadpool_info():
...
+
+    Parameters
+        ----------
+        extra_info : bool
+
+            Include extra fields which requires more intrusive actions to
+            obtain.
+
+            - "api_scope": When setting the number of threads, what is affected.
+                           Possible values are "process", "current_thread".
```
```suggestion
    Parameters
    ----------
    extra_info : bool
        Include extra fields which requires more intrusive actions to obtain.
 
        - "api_scope": When setting the number of threads, what is affected.
          Possible values are "process", "current_thread".
```

threadpoolctl.py
```diff
@@ -824,9 +939,20 @@ def _from_controllers(cls, lib_controllers):
...
-    def info(self):
-        """Return lib_controllers info as a list of dicts"""
-        return [lib_controller.info() for lib_controller in self.lib_controllers]
+    def info(self, extra_info: bool = False):
+        """Return lib_controllers info as a list of dicts.
+
+        Parameters
+        ----------
+        extra_info : bool
+
```
```suggestion
```

threadpoolctl.py
ogrisel: Or alternatively `_NumThreadsScope`.

threadpoolctl.py
```diff
@@ -69,6 +71,96 @@ class _dl_phdr_info(ctypes.Structure):
...
+    The function works by changing the number of threads in loaded controllers,
+    which can be a process-wide change. As such, it is not always thread-safe.
+    An attempt will be made to restore all settings to their previous state,
+    but the result may be subtly different, e.g. if "unset" has different
+    semantics than "set to the default returned value".
+    """
+    if os.cpu_count() == 1 or (
+        hasattr(os, "process_cpu_count") and os.process_cpu_count() == 1
+    ):
+        raise RuntimeError("Cannot determine API meaning if only one core is available")
```
ogrisel: Shouldn't this rather this be a warning? We can always return "UNKNOWN" in that case.

I think `python -m threadpoolctl` should not fail on single core hosts.

threadpoolctl.py
```diff
@@ -563,8 +668,18 @@ def threadpool_info():
...
+    Parameters
+        ----------
+        extra_info : bool
+
+            Include extra fields which requires more intrusive actions to
+            obtain.
+
+            - "api_scope": When setting the number of threads, what is affected.
+                           Possible values are "process", "current_thread".
     """
```
ogrisel, line 683: When `extra_info` is `False` I would be in favor of retrieving the expected info from a statically hardcoded data based of known semantics for common BLAS and OpenMP implementation and return UNKNOWN otherwise.

Then we can have tests to check that `threadpool_info(extra_info=True)` always returns the same as `threadpool_info(extra_info=False)` on all the environments tested by our CI.

ogrisel: If we do so, we should probably rename `extra_info` to `inspect_scope` or something like that.

threadpoolctl.py
itamarst: Thinking out loud... also relevant to the "should we hardcode this if introspection isn't available" question...

My current thought is that introspection is a useful debugging tool, but not something that would actually be _used_. Whatever it does, the library is _stuck_ with it. And so I think it's OK if it's just opt-in, mostly for CLI users doing bug reports or when doing performance debugging.

And so something simpler and minimalist seems sufficient.

I may be wrong, it may be that knowing which it is will be helpful. And if so we can maybe go for more elaborate approaches like subprocess later on.

threadpoolctl.py
itamarst: See my comments about subprocess implementation elsewhere, I think this is probably unnecessary.

EDIT: https://github.com/joblib/threadpoolctl/pull/213#discussion_r3506291561

itamarst: Basically I am thinking of this for now at least mostly as debugging and bug report tool, rather than something the API will use.

threadpoolctl.py
itamarst: Also that message is wrong; it can be determined. So just deleting that whole block.

itamarst:
```

++ /home/runner/miniconda3/bin/conda shell.posix reactivate
  File "/home/runner/miniconda3/bin/conda", line 12, in <module>
    from conda.cli import main
ModuleNotFoundError: No module named 'conda'
```

That's a fun error.

Probably due to https://github.com/conda/conda/issues/16327

itamarst:
I addressed or replied to all review comments. CI is broken for now, will try again later.

itamarst:
I wonder why a filter of `{'internal_api': 'openblas', 'threading_layer': 'pthreads'}` is returning a BLISController... will get to that next week, I guess.

threadpoolctl.py
ogrisel: I think we should make that explicit in the public API because otherwise libraries will start depending on it.

itamarst: I renamed `extra_info` to `debugging_info`, and added a bit more prose explanation about it being unreliable.

itamarst:
> I wonder why a filter of `{'internal_api': 'openblas', 'threading_layer': 'pthreads'}` is returning a BLISController... will get to that next week, I guess.

Ah, I misunderstood the logic, it's `OR` not `AND`. Will fix.

itamarst:
Ready for another round of review, I think.

tests/test_api_introspection.py
```diff
@@ -0,0 +1,101 @@
...
+@pytest.mark.skipif(
+    sys.platform != "linux", reason="Non-Linux OpenMP might be different"
+)
+@pytest.mark.parametrize(
+    ["select_filter", "expected_thread_limit_scope", "extra_check"],
+    [
+        (
+            {"internal_api": "openblas"},
+            _ThreadLimitScope.PROCESS,
+            lambda lib: lib.threading_layer == "pthreads",
```
ogrisel, line 78: Maybe add a reason why we skip the extra check for the openmp threading layer.

ogrisel (approved):
Let's add a changelog entry for this. Otherwise, LGTM.

Any opinion @jeremiedbb?

ogrisel:
The CI failures seem to be unrelated I think. If so we can fix them in a dedicated PR.

itamarst:
> The CI failures seem to be unrelated I think. If so we can fix them in a dedicated PR.

I believe there's an outstanding PR that fixes one of them, yes.

itamarst:
Added changelog and explanatory comment.

ogrisel:
cc @tomMoral in case you are looking for a quick review at night :)

CHANGES.md
```diff
@@ -1,6 +1,10 @@
 3.7.0 (TBD)
 ===========
 
+- Added the ability to check whether a limiting API
+  affects just the current thread or the whole process. Mainly aimed at debugging and diagnostics, and somewhat unreliable, it is therefore enabled by default only for command-line usage.
+  https://github.com/joblib/threadpoolctl/pull/213
```
```suggestion
- Added the ability to check whether a limiting API affects just the current
  thread or the whole process. Mainly aimed at debugging and diagnostics, and
  somewhat unreliable, it is therefore enabled by default only for command-line
  usage.
  https://github.com/joblib/threadpoolctl/pull/213
```

threadpoolctl.py
```diff
@@ -69,6 +71,91 @@ class _dl_phdr_info(ctypes.Structure):
     _RTLD_NOLOAD = ctypes.DEFAULT_MODE
 
 
+class _ThreadLimitScope(Enum):
```
tomMoral: This feels quite complex for a three string returns that are setup only once in the code.
I tend to prefer bare string, that is more coherent with the rest of the code (but probably a matter of taste).

threadpoolctl.py
```diff
@@ -69,6 +71,91 @@ class _dl_phdr_info(ctypes.Structure):
...
+        # determine how the API works. We try to not reach that point here, but
+        # you can imagine a thread pool implementation that is aware of
+        # cgroups, in which case a Docker container limited to one core will
+        # pass the safety check at the start of the function. Perhaps
+        # cpu_count() from loky should be moved into this package...
+        if thread_result != [expected]:
+            return _ThreadLimitScope.UNKNOWN
+
+        # Now, check this thread:
+        if get_n_threads() == expected:
```
tomMoral, line 143: Would be interesting on the thread local setup to check that the return of get_n_theads is also the previous one, as there could be border effects of initiating a new thread?

But maybe that's plan in follow ups

tomMoral (approved):
This looks like a super cool feature that will help understand the scope, get better reports.

One potential caveat is with the initialized vs non initialized library. Maybe worth having a way to run this in a fresh subprocess/check effects in different threads?

threadpoolctl.py
itamarst: If this ever grows type checking, the alternatives involves something like:

```python
type _ThreadLimitScope  = Literal["current_thread", "process", "unknown"]
```

And then once you add all the explanatory comments in it is I guess shorter, but not by much.

I am happy to change it though if you prefer.

threadpoolctl.py
itamarst: I'll do that.

itamarst: Actually, isn't this already implemented in `elif` clause? Or maybe I'm not understanding what you're asking for.

itamarst:
OK, anything else left to do before this can be merged?

itamarst:
> One potential caveat is with the initialized vs non initialized library. Maybe worth having a way to run this in a fresh subprocess/check effects in different threads?

There's nothing stopping users from calling it themselves, later on, via the Python API?

@tomMoral if you're OK with this, could you merge? That would unblock the next PR in the queue.

itamarst:
Just checking again, could we get this merged? Then I can move on to the next PR.

ogrisel:
Merged. I will review the second part.

Issue #216, opened 2026-07-01
Use thread-local APIs when possible


itamarst:
Given the existence in the wild (e.g. in scikit-learn) of nested Python thread pools, being able to set limits per-thread is that much more important for reliable oversaturation prevention.

Some libraries provide both process-wide and thread-local settings:

* MKL: https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2023-0/mkl-set-num-threads-local.html
* BLIS: https://github.com/flame/blis/blob/master/docs/Multithreading.md#locally-at-runtime

In practice BLIS is probably not viable.

Long-term, I think there should be a ecosystem-wide push to provide thread-local APIs.

In the short term, exposing MKL at least would be helpful for #215.

**Update:** After some discussion of the downstream use case of Python thread pools  (#215), conclusion was that rather than it being useful to have thread-local scoped limiting APIs, they should in fact be the default.

itamarst:
My first thought was just adding it for MKL, but given #211 one could actually do this dynamically for additional APIs, if sufficiently motivated, so might wait for that.

UPDATE: Maybe not actually useful to do this dynamically? You're still stuck with whatever the API gives you, so nested Python thread pools are still a problem if all you have is process-level setting scope. So maybe I'll revert to initial plan.

itamarst:
Did most of the work in a branch, but will wait until #214 is merged so I can write a test.

ogrisel:
> In practice BLIS is probably not viable.

I agree: I took at the BLIS API, and it seems that using it would require callers to be recompiled to use non-standard BLAS function equivalent to pass the locally configured runtime object as extra argument:

https://github.com/flame/blis/blob/master/docs/Multithreading.md#locally-at-runtime-using-the-expert-interfaces

ogrisel:
For the record, OpenBLAS also has a `openblas_set_num_threads_local` function, but it is not documented, and after [looking at the code](https://github.com/search?q=repo%3AOpenMathLib%2FOpenBLAS%20openblas_set_num_threads_local&type=code) I doubt that this is equivalent to what `mkl_set_num_threads_local`. This function does not seem to be tested either.

If needed, we could try to contribute to OpenBLAS to actually make this function use a thread-local configuration, with similar semantics as the MKL API but that sounds complex to get right.

itamarst:
[...]

Issue #215, opened 2026-06-30
API for setting limits for use by Python thread pools


itamarst:
See #208 for discussion of why the current API is insufficient.

Here is a demonstration of the problem as far as the current API goes, will eventually become a unit test for new API:

```python
from sys import thread_info
from threading import Thread, current_thread

import sklearn
from threadpoolctl import threadpool_info, threadpool_limits

def assert_limit_is(limit):
    print("Checking", current_thread())
    for i in threadpool_info():
        assert i["num_threads"] == limit
    print("OK")


threadpool_limits(limits=2)
assert_limit_is(2)

t = Thread(target=lambda: assert_limit_is(2))
t.start()
t.join()
```

When run:

```
Checking <_MainThread(MainThread, started 128633514360960)>
OK
Checking <Thread(Thread-1, started 128632355026624)>
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib/python3.9/threading.py", line 980, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.9/threading.py", line 917, in run
    self._target(*self._args, **self._kwargs)
  File "/home/itamarst/devel/threadpoolctl/theproblem.py", line 17, in <lambda>
    t = Thread(target=lambda: assert_limit_is(2))
  File "/home/itamarst/devel/threadpoolctl/theproblem.py", line 10, in assert_limit_is
    assert i["num_threads"] == limit
AssertionError
```

## Thinking through the design space

Limiting APIs might affect only the current thread, or all threads in the process. If the former, we need to call the API in every thread in the thread pool; if the latter, we need to call it before starting the thread pool in the thread that starts the pool.
 
If every library has a single exposed API, there's really no difference between the logic in the current API `threadpool_limits()` and what you'd do in a thread pool thread. You don't have a choice, you just have the one API!

However, sometimes there is a choice. For MKL there will be a choice between "apply to all threads" vs "apply just to this thread". If there _is_ a choice for a library:

* In the main thread, you don't really need to call anything for the library, if the presumed usage pattern is that all work is done in the new thread pool. In fact you _don't_ want to call process-wide APIs because of nested thread pools.
* In the threads in the thread pool, you want to call the thread-local API.

## Proposed solutions

Noted that for all options, nested Python thread pools are gonna be half-broken given some limiting APIs are process-wide. But will kinda-sorta work if all threads take the same amount of time to run... I have an idea for a completely different approach to deal with that though.

[...]

PR #218, opened 2026-07-07
Documentation for proposed API to cover Python thread pool use case

itamarst:
@ogrisel some questions:

1. Does this seem reasonable from a user perspective?
2. Are there other use cases I'm not thinking of? Writing the docs brought up two, so maybe there are more.

Actual coding depends on #213 and #216.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,17 +172,18 @@ usage section.
 
 Limiting thread pool size in controlled libraries requires a two-step process.
 Importantly, each Python thread must call a method to limit controlled libraries
-in that thread. With Python's `multiprocessing.pool.ThreadPool`, you can do so
-by passing in an initializer function that will get called on thread startup.
+in that thread. With Python's `concurrent.futures.ThreadPoolExecutor`, you can
+do so by passing in an initializer function that will get called on thread
+startup.
 
 ```python
 from threadpoolctl import LimiterForPythonThreads
-from multiprocessing.pool import ThreadPool
+from concurrent.futures import ThreadPoolExecutor
 
 with limit_for_python_threads(limits=1, user_api='blas') as limiter:
     # Make sure each Python thread calls limiter.limit_in_pythread(). If you're
     # using another thread pool class, you will need to do some other way.
-    with ThreadPool(4, initializer=limiter.limit_in_pythread) as pool:
+    with ThreadPoolExecutor(4, initializer=limiter.limit_in_pythread) as pool:
         # ... run some BLAS-using code in the thread pool ...
 ```
 
@@ -196,12 +197,12 @@ from threadpoolctl import ThreadpoolController
 CONTROLLER = ThreadpoolController()
 
 with CONTROLLER.limit_for_python_threads(limits=1) as limiter:
-    with ThreadPool(4, initializer=limiter.limit_in_pythread) as pool:
+    with ThreadPoolExecutor(4, initializer=limiter.limit_in_pythread) as pool:
         # ... run some BLAS-using code in the thread pool ...
 
 # Later...
 with CONTROLLER.limit_for_python_threads(limits=2) as limiter:
-    with ThreadPool(2, initializer=limiter.limit_in_pythread) as pool:
+    with ThreadPoolExecutor(2, initializer=limiter.limit_in_pythread) as pool:
         # ... run some BLAS-using code in the thread pool ...
 ```
 
@@ -216,7 +217,7 @@ more complex: you need to set the limits each time you switch back and forth.
 Let's say your computer has 4 cores, and you're using some OpenMP API.
 
 ```python
-POOL = ThreadPool(4)
+POOL = ThreadPoolExecutor(4)
 CONTROLLER = ThreadpoolController()
 
 # 1. Run some work in a Python thread pool, which then runs in OpenMP.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -152,6 +152,104 @@ The state of these libraries is also accessible through the object oriented API:
...
 ```
 
+## Usage when using Python threads: Restricting Controlled Library Thread Pool Sizes
+
+There a two scenarios in which you might want to use `threadpoolctl`; each
+requires you to use different APIs.
+
+1. You will be parallelizing work using a Python thread pool, and your goal is
+   therefore to limit controlled libraries' thread pool sizes when used from
+   Python threads.
```
ogrisel:
```suggestion
   therefore to limit controlled libraries' thread pool sizes when
   concurrently called from Python threads. This case is a bit more
   complex to handle properly and requires a bit more verbose code.
```

README.md
```diff
@@ -152,6 +152,104 @@ The state of these libraries is also accessible through the object oriented API:
...
+
+This section will cover the former case, and the latter is covered in the next
+usage section.
+
+### Setting the Maximum Size of Thread-Pools, When Python Thread Pools Are Used
+
+Limiting thread pool size in controlled libraries requires a two-step process.
+Importantly, each Python thread must call a method to limit controlled libraries
+in that thread. With Python's `multiprocessing.pool.ThreadPool`, you can do so
+by passing in an initializer function that will get called on thread startup.
```
ogrisel: `multiprocessing.pool.ThreadPool` feels a bit old-school. I would rather use the slightly more modern `concurrent.futures.ThreadPoolExecutor` alternative in documentation written in 2026.

This class also exposes an `initializer` argument.

README.md
```diff
@@ -152,6 +152,104 @@ The state of these libraries is also accessible through the object oriented API:
...
+with CONTROLLER.limit_for_python_threads(limits=1) as limiter:
+    with ThreadPool(4, initializer=limiter.limit_in_pythread) as pool:
+        # ... run some BLAS-using code in the thread pool ...
+
+# Later...
+with CONTROLLER.limit_for_python_threads(limits=2) as limiter:
+    with ThreadPool(2, initializer=limiter.limit_in_pythread) as pool:
+        # ... run some BLAS-using code in the thread pool ...
+```
+
```
ogrisel, line 271: Maybe we could warn users that changing the thread pool sizes is expected to have some overhead and should not be done from within an otherwise fast iterating loop.

README.md
```diff
@@ -152,6 +152,104 @@ The state of these libraries is also accessible through the object oriented API:
...
+    def limit_then_do_work2(*args, **kwargs):
+        limiter.limit_in_pythread()
+        return do_even_more_real_work_with_openmp(*args, **kwargs)
+
+    results3 = POOL.map(limit_then_do_work2, results2)
+```
+
+## Usage for no Python threads: Restricting Controlled Library Thread Pool Sizes
+
+This covers APIs to use when you don't expect to use Python thread pools to parallelize work.
```
ogrisel: I think we should start with this simple case (process-wide API) before explaining how to control threadpool libraries called from Python threads.

README.md
```diff
@@ -184,11 +282,17 @@ however not act on libraries loaded after the instantiation of the
...
+
 ### Restricting the limits to the scope of a function
 
 `threadpool_limits` and `ThreadpoolController` can also be used as decorators to set
 the maximum number of threads used by the supported libraries at a function level. The
-decorators are accessible through their `wrap` method:
+decorators are accessible through their `wrap` method.
+
+This should only be used for functions you expect to be called from the main
+thread of a process, without the use of any Python thread pools.
```
ogrisel: Wouldn't this approach also work if you passed such a decorated callable to the `POOL.map` as an alternative to what is presented in "Switching Back And Forth Between Main Thread and Python Threads"?

README.md
```diff
@@ -152,6 +152,104 @@ The state of these libraries is also accessible through the object oriented API:
...
+## Usage when using Python threads: Restricting Controlled Library Thread Pool Sizes
+
+There a two scenarios in which you might want to use `threadpoolctl`; each
+requires you to use different APIs.
+
+1. You will be parallelizing work using a Python thread pool, and your goal is
+   therefore to limit controlled libraries' thread pool sizes when used from
+   Python threads.
+2. You do not expect to use any Python threads, so all the work will be started
+   directly from the main thread in the process.
```
ogrisel: Let's start with this and maybe make it explicit:


```suggestion
   directly from the main thread in the process. This is a simple case
   where we can globally set thread limits.
```

ogrisel (commented):
Thanks for the PR. Here is some feedback.

README.md
itamarst: It's probably not that expensive compared to anything you're likely to run that uses BLAS or OpenMP...

README.md
itamarst: If it's called from a thread, then it's not equivalent to the per-thread API, e.g. it won't know to call MKL's thread-specific limiting API and will instead want to use MKL's process-wide limiting API.

itamarst:
Since it sounded like the API design itself is probably OK, I will probably close this draft PR for now and re-open when there's code to show.

ogrisel:
@itamarst Actually, I thought a bit more about this and I think we don't necessarily need to change the existing code but rather document how to use it correctly in a Python-level threaded app as explained in https://github.com/joblib/threadpoolctl/issues/215#issuecomment-4922756383.

PR #189, opened 2025-03-11
ENH Support path length > MAX_PATH on Windows

jeremiedbb:
Closes https://github.com/joblib/threadpoolctl/issues/181

- limit extended to 10 * MAX_PATH (2600)
- kept a hard limit for security reasons
- A warning is raised if a library has a path too long, and this lib is ignored

PR #219, opened 2026-07-09
FIX intermittent Windows module enumeration failure

### LINKED ISSUE #217: ThreadpoolController() raises an error intermittently with conda-forge OpenCV on Windows
Hello

I am getting an  intermittent ``OSError``  when a ``ThreadpoolController``  is created soon after importing the conda-forge OpenCV package on Windows.  This reproduces the problem reliably on my PC:

```python
import cv2

for _ in range(1000):
    ThreadpoolController()
```

This is the stack trace: 

```python
  File "C:\Temp\threadpoolctl_test.py", line 5, in <module>
    ThreadpoolController()
    ~~~~~~~~~~~~~~~~~~~~^^
  File "C:\Users\dugalh\miniconda3\envs\py313-geospatial\Lib\site-packages\threadpoolctl.py", line 818, in __init__
    self._load_libraries()
    ~~~~~~~~~~~~~~~~~~~~^^
  File "C:\Users\dugalh\miniconda3\envs\py313-geospatial\Lib\site-packages\threadpoolctl.py", line 972, in _load_libraries
    self._find_libraries_with_enum_process_module_ex()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "C:\Users\dugalh\miniconda3\envs\py313-geospatial\Lib\site-packages\threadpoolctl.py", line 1099, in _find_libraries_with_enum_process_module_ex
    raise OSError("GetModuleFileNameEx failed")
OSError: GetModuleFileNameEx failed
```

I have done some debugging and think the problem occurs when DLL(s) are unloaded between the ``EnumProcessModulesEx()`` and ``GetModuleFileNameExW()`` calls in ``ThreadpoolController._find_libraries_with_enum_process_module_ex()``.  Then ``GetModuleFileNameExW()``  fails when it is passed handle(s) of unloaded DLL(s).   

In addition to the possibility of DLL(s) being unloaded between these calls, the ``EnumProcessModulesEx()`` [docs](https://learn.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-enumprocessmodules) say "If the module list in the target process is corrupted or not yet initialized, or if the module list changes during the function call as a result of DLLs being loaded or unloaded, EnumProcessModules may fail or return incorrect information."

An alternative to ``EnumProcessModulesEx()`` / ``GetModuleFileNameExW()``  is [``CreateToolhelp32Snapshot()``](https://learn.microsoft.com/en-us/windows/win32/api/tlhelp32/nf-tlhelp32-createtoolhelp32snapshot).  E.g. this function returns the currently loaded DLLs:

```python
import ctypes

def _find_libraries_with_create_tool_help_snapshot():
    from ctypes import wintypes

    class MODULEENTRY32W(ctypes.Structure):
        # ctypes definition of:
        # https://learn.microsoft.com/en-us/windows/win32/api/tlhelp32/ns-tlhelp32-moduleentry32w
        _fields_ = [
            ('dwSize', wintypes.DWORD),
            ('th32ModuleID', wintypes.DWORD),
            ('th32ProcessID', wintypes.DWORD),
            ('GlblcntUsage', wintypes.DWORD),
            ('ProccntUsage', wintypes.DWORD),
            ('modBaseAddr', ctypes.POINTER(wintypes.BYTE)),
            ('modBaseSize', wintypes.DWORD),
            ('hModule', wintypes.HMODULE),
            ('szModule', wintypes.WCHAR * 256),
            ('szExePath', wintypes.WCHAR * wintypes.MAX_PATH),
        ]

    TH32CS_SNAPMODULE = 0x8
    TH32CS_SNAPMODULE32 = 0x10
    ERROR_BAD_LENGTH = 0x18
    INVALID_HANDLE_VALUE = wintypes.HANDLE(-1).value
    ERROR_NO_MORE_FILES = 0x12

    kernel_32 = ctypes.WinDLL("kernel32.dll", use_last_error=True)

    # ctype function definitions
    CreateToolhelp32Snapshot = kernel_32.CreateToolhelp32Snapshot
    CreateToolhelp32Snapshot.argtypes = [wintypes.DWORD, wintypes.DWORD]
    CreateToolhelp32Snapshot.restype = wintypes.HANDLE

    Module32FirstW = kernel_32.Module32FirstW
    Module32FirstW.argtypes = [wintypes.HANDLE, ctypes.POINTER(MODULEENTRY32W)]
    Module32FirstW.restype = wintypes.BOOL

    Module32NextW = kernel_32.Module32NextW
    Module32NextW.argtypes = [wintypes.HANDLE, ctypes.POINTER(MODULEENTRY32W)]
    Module32NextW.restype = wintypes.BOOL

    CloseHandle = kernel_32.CloseHandle
    CloseHandle.argtypes = [wintypes.HANDLE]
    CloseHandle.restype = wintypes.BOOL

    # take a snapshot of loaded libraries
    while True:
        snap_handle = CreateToolhelp32Snapshot(TH32CS_SNAPMODULE | TH32CS_SNAPMODULE32, 0)
        if snap_handle == INVALID_HANDLE_VALUE:
            err = ctypes.get_last_error()
            if err == ERROR_BAD_LENGTH:
                # retry until it succeeds
                continue
            msg = ctypes.FormatError(err).strip()
            raise ctypes.WinError(err, f'CreateToolhelp32Snapshot failed: {msg}')
        break

    libraries = []
    try:
        # iterate through the snapshot, retrieving library paths
        lib_entry = MODULEENTRY32W()
        lib_entry.dwSize = ctypes.sizeof(MODULEENTRY32W)
        if not (res := Module32FirstW(snap_handle, ctypes.byref(lib_entry))):
            err = ctypes.get_last_error()
            msg = ctypes.FormatError(err).strip()
            raise ctypes.WinError(err, f'Module32FirstW failed: {msg}')

        while res:
            libraries.append(lib_entry.szExePath)
            res = Module32NextW(snap_handle, ctypes.byref(lib_entry))

        if err := ctypes.get_last_error() != ERROR_NO_MORE_FILES:
            msg = ctypes.FormatError(err).strip()
            raise ctypes.WinError(err, f'Module32NextW failed: {msg}')
    finally:
        CloseHandle(snap_handle)

    return libraries
```

I think this is more robust to DLLs being loaded/unloaded, but it fails if any of the loaded DLL path lengths are longer than ``MAX_PATH``.  That is not really an issue for me, but it could be for other users?

ogrisel:
Fix #217

## Summary

- Fix intermittent `OSError("GetModuleFileNameEx failed")` on Windows when DLLs are loaded or unloaded during library discovery (reported with conda-forge OpenCV).
- Replace the two-phase `EnumProcessModulesEx` + `GetModuleFileNameExW` scan with a snapshot-first approach using `CreateToolhelp32Snapshot`, keeping long-path support via `GetModuleFileNameExW` and graceful per-module fallbacks.
- Add Windows integration tests for library paths longer than `MAX_PATH`, inspired by the local repro documented in https://github.com/joblib/threadpoolctl/pull/189#issuecomment-2714235916.

Closes https://github.com/joblib/threadpoolctl/issues/217

## Test plan

- [x] Windows CI (`py311_conda_forge_openblas`): new tests `test_windows_library_path_longer_than_max_path` and `test_windows_library_path_exceeds_internal_limit` pass
- [x] Windows CI: existing BLAS/OpenMP discovery tests still pass
- [x] Manual repro from issue #217 with conda-forge OpenCV:
  ```python
  import cv2

  for _ in range(1000):
      ThreadpoolController()
  ```

## AI assistance disclosure

This pull request was prepared with assistance from an AI coding agent (Cursor / Claude).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,8 +3,9 @@
 
 - Fixed an intermittent `OSError` on Windows when DLLs are loaded or unloaded
   concurrently during library discovery (for example when importing conda-forge
-  OpenCV). Windows module enumeration now uses a snapshot-first approach with
-  graceful per-module fallbacks.
+  OpenCV). On Python 3.14+, discovery uses `ctypes.util.dllist` when available
+  and falls back to a Toolhelp snapshot if that raises. Older Pythons use the
+  snapshot enumerator, with graceful per-module fallbacks.
   https://github.com/joblib/threadpoolctl/pull/219
 
 - Added the ability to check whether a limiting API affects just the current
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,15 @@
+3.7.0 (TBD)
+===========
+
+- Fixed an intermittent `OSError` on Windows when DLLs are loaded or unloaded
+  concurrently during library discovery (for example when importing conda-forge
+  OpenCV). Windows module enumeration now uses a snapshot-first approach with
+  graceful per-module fallbacks.
+  https://github.com/joblib/threadpoolctl/issues/217
+
+- Added Windows integration tests for library paths longer than `MAX_PATH`.
+  https://github.com/joblib/threadpoolctl/pull/189
+
 3.6.0 (2025-03-13)
 ==================
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,9 +7,6 @@
   graceful per-module fallbacks.
   https://github.com/joblib/threadpoolctl/issues/217
 
-- Added Windows integration tests for library paths longer than `MAX_PATH`.
-  https://github.com/joblib/threadpoolctl/pull/189
-
 3.6.0 (2025-03-13)
 ==================
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,9 +3,9 @@
 
 - Fixed an intermittent `OSError` on Windows when DLLs are loaded or unloaded
   concurrently during library discovery (for example when importing conda-forge
-  OpenCV). On Python 3.14+, discovery uses `ctypes.util.dllist` when available
-  and falls back to a Toolhelp snapshot if that raises. Older Pythons use the
-  snapshot enumerator, with graceful per-module fallbacks.
+  OpenCV). On Python 3.14+, discovery uses `ctypes.util.dllist` when available.
+  Older Pythons use a Toolhelp snapshot enumerator, with graceful per-module
+  fallbacks.
   https://github.com/joblib/threadpoolctl/pull/219
 
 - Added the ability to check whether a limiting API affects just the current
--- a/continuous_integration/check_no_test_skipped.py
+++ b/continuous_integration/check_no_test_skipped.py
@@ -36,7 +36,12 @@ print("\n------------------------------------------------------------------\n")
 # List of tests that we don't want to fail the CI if they are skipped in
 # every job. This is useful for tests that depend on specific versions of
 # numpy or scipy and we don't want to pin old versions of these libraries.
-SAFE_SKIPPED_TESTS = ["test_multiple_shipped_openblas"]
+SAFE_SKIPPED_TESTS = [
+    "test_multiple_shipped_openblas",
+    # Requires Windows with shipped OpenBLAS (e.g. py311_conda_forge_openblas).
+    "test_windows_library_path_longer_than_max_path",
+    "test_windows_library_path_exceeds_internal_limit",
+]
 
 fail = False
 for test, skipped in always_skipped.items():
--- a/continuous_integration/check_no_test_skipped.py
+++ b/continuous_integration/check_no_test_skipped.py
@@ -36,12 +36,7 @@ print("\n------------------------------------------------------------------\n")
 # List of tests that we don't want to fail the CI if they are skipped in
 # every job. This is useful for tests that depend on specific versions of
 # numpy or scipy and we don't want to pin old versions of these libraries.
-SAFE_SKIPPED_TESTS = [
-    "test_multiple_shipped_openblas",
-    # Requires Windows with shipped OpenBLAS (e.g. py311_conda_forge_openblas).
-    "test_windows_library_path_longer_than_max_path",
-    "test_windows_library_path_exceeds_internal_limit",
-]
+SAFE_SKIPPED_TESTS = ["test_multiple_shipped_openblas"]
 
 fail = False
 for test, skipped in always_skipped.items():
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -835,7 +835,12 @@ def test_windows_library_path_longer_than_max_path(tmp_path):
         tmp_path, "libopenblas_long_path_test.dll", min_length=261
     )
     extended_path = os.path.abspath(str(long_path))
+    # Win32 APIs reject paths longer than MAX_PATH unless they use the
+    # extended-length prefix ``\\?\``. Skip rewriting if abspath already
+    # returned that form.
     if not extended_path.startswith("\\\\?\\"):
+        # UNC paths (``\\server\share\...``) need ``\\?\UNC\server\share\...``
+        # rather than ``\\?\\server\...``.
         if extended_path.startswith("\\\\"):
             extended_path = "\\\\?\\UNC\\" + extended_path[2:]
         else:
@@ -844,6 +849,9 @@ def test_windows_library_path_longer_than_max_path(tmp_path):
     ctypes.CDLL(extended_path)
 
     expected_path = os.path.abspath(str(long_path))
+    # Compare using the canonical path without the ``\\?\`` prefix.
+    # ``\\?\C:\...`` becomes ``C:\...``; ``\\?\UNC\server\share`` becomes
+    # ``\\server\share``.
     if expected_path.startswith("\\\\?\\"):
         expected_path = expected_path[4:]
         if expected_path.startswith("UNC\\"):
@@ -856,6 +864,8 @@ def test_windows_library_path_longer_than_max_path(tmp_path):
     normalized_filepaths = []
     for info in long_path_entries:
         filepath = info["filepath"]
+        # Same prefix stripping as expected_path so discovery results match
+        # regardless of whether GetModuleFileNameExW kept ``\\?\``.
         if filepath.startswith("\\\\?\\"):
             filepath = filepath[4:]
             if filepath.startswith("UNC\\"):
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -7,7 +7,7 @@ import sys
 
 import threadpoolctl
 from threadpoolctl import threadpool_limits, threadpool_info
-from threadpoolctl import ThreadpoolController, _realpath
+from threadpoolctl import ThreadpoolController
 from threadpoolctl import _ALL_PREFIXES, _ALL_USER_APIS
 
 from .utils import cython_extensions_compiled
@@ -15,7 +15,10 @@ from .utils import check_nested_prange_blas
 from .utils import libopenblas_paths
 from .utils import copy_and_load_dll
 from .utils import get_openblas_dll_path
+from .utils import LONG_PATH_OPENBLAS_DLL
 from .utils import make_long_windows_path
+from .utils import normalize_windows_path
+from .utils import TRUNCATED_PATH_OPENBLAS_DLL
 from .utils import scipy
 from .utils import threadpool_info_from_subprocess
 from .utils import select
@@ -820,17 +823,17 @@ def test_windows_library_path_longer_than_max_path(tmp_path):
     if src_dll is None:
         pytest.skip("Requires OpenBLAS on Windows")
 
-    dll_name = os.path.basename(src_dll)
-    long_path = make_long_windows_path(tmp_path, dll_name, min_length=261)
+    long_path = make_long_windows_path(tmp_path, LONG_PATH_OPENBLAS_DLL, min_length=261)
     copy_and_load_dll(src_dll, long_path)
 
-    expected_path = _realpath(str(long_path))
+    expected_path = normalize_windows_path(long_path)
     openblas_info = ThreadpoolController().select(internal_api="openblas").info()
 
     long_path_entries = [info for info in openblas_info if len(info["filepath"]) > 260]
     assert len(long_path_entries) >= 1
     assert any(
-        _realpath(info["filepath"]) == expected_path for info in long_path_entries
+        normalize_windows_path(info["filepath"]) == expected_path
+        for info in long_path_entries
     )
 
 
@@ -847,13 +850,18 @@ def test_windows_library_path_exceeds_internal_limit(tmp_path, monkeypatch):
 
     monkeypatch.setattr(threadpoolctl, "_WINDOWS_MAX_LIBRARY_PATH_LENGTH", 300)
 
-    dll_name = "libopenblas_path_too_long.dll"
-    long_path = make_long_windows_path(tmp_path, dll_name, min_length=301)
+    long_path = make_long_windows_path(
+        tmp_path, TRUNCATED_PATH_OPENBLAS_DLL, min_length=400
+    )
     copy_and_load_dll(src_dll, long_path)
 
-    expected_path = _realpath(str(long_path))
+    expected_path = normalize_windows_path(long_path)
     with pytest.warns(RuntimeWarning, match="path too long"):
         info = ThreadpoolController().info()
 
-    filepaths = {_realpath(entry["filepath"]) for entry in info if "filepath" in entry}
+    filepaths = {
+        normalize_windows_path(entry["filepath"])
+        for entry in info
+        if "filepath" in entry
+    }
     assert expected_path not in filepaths
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -15,6 +15,7 @@ from threadpoolctl import _ALL_PREFIXES, _ALL_USER_APIS
 from .utils import cython_extensions_compiled
 from .utils import check_nested_prange_blas
 from .utils import libopenblas_paths
+from .utils import get_openblas_dll_path
 from .utils import make_long_windows_path
 from .utils import scipy
 from .utils import threadpool_info_from_subprocess
@@ -816,10 +817,10 @@ def test_windows_library_path_longer_than_max_path(tmp_path):
     Regression test inspired by the local repro in
     https://github.com/joblib/threadpoolctl/pull/189#issuecomment-2714235916
     """
-    if not libopenblas_paths:
-        pytest.skip("Requires numpy with shipped OpenBLAS on Windows")
+    src_dll = get_openblas_dll_path()
+    if src_dll is None:
+        pytest.skip("Requires OpenBLAS on Windows")
 
-    src_dll = next(iter(libopenblas_paths))
     dll_name = os.path.basename(src_dll)
     long_path = make_long_windows_path(tmp_path, dll_name, min_length=261)
     shutil.copy2(src_dll, long_path)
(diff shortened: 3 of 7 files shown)


continuous_integration/check_no_test_skipped.py
```diff
@@ -36,7 +36,12 @@
 # List of tests that we don't want to fail the CI if they are skipped in
 # every job. This is useful for tests that depend on specific versions of
 # numpy or scipy and we don't want to pin old versions of these libraries.
-SAFE_SKIPPED_TESTS = ["test_multiple_shipped_openblas"]
+SAFE_SKIPPED_TESTS = [
+    "test_multiple_shipped_openblas",
+    # Requires Windows with shipped OpenBLAS (e.g. py311_conda_forge_openblas).
+    "test_windows_library_path_longer_than_max_path",
+    "test_windows_library_path_exceeds_internal_limit",
```
ogrisel: Since we expect the `py311_conda_forge_openblas` to run, I think we don't need this::

 
```suggestion
```

tests/test_threadpoolctl.py
```diff
@@ -792,3 +796,67 @@ def test_custom_controller():
         assert mylib_controller.num_threads == 1
 
     assert ThreadpoolController().info() == original_info
+
+
+@pytest.mark.skipif(sys.platform != "win32", reason="Windows-only test")
+def test_threadpool_controller_repeated_init_on_windows():
```
ogrisel: We could run this test on all platforms and just mention in the docstring that this is a non-regression test for a Windows-specific problem.

tests/utils.py
```diff
@@ -17,6 +18,13 @@
     np.dot(np.ones(1000), np.ones(1000))
 
     libopenblas_patterns.append(os.path.join(np.__path__[0], ".libs", "libopenblas*"))
+    if sys.platform == "win32":
+        libopenblas_patterns.append(
+            os.path.join(np.__path__[0], "numpy.libs", "libscipy_openblas*.dll")
+        )
+        libopenblas_patterns.append(
+            os.path.join(np.__path__[0], "numpy.libs", "libopenblas*.dll")
```
ogrisel: This seems redundant.

tests/utils.py
```diff
@@ -17,6 +18,13 @@
     np.dot(np.ones(1000), np.ones(1000))
 
     libopenblas_patterns.append(os.path.join(np.__path__[0], ".libs", "libopenblas*"))
+    if sys.platform == "win32":
+        libopenblas_patterns.append(
+            os.path.join(np.__path__[0], "numpy.libs", "libscipy_openblas*.dll")
+        )
```
ogrisel: Why the `sys.platform == "win32"` condition?

CHANGES.md
```diff
@@ -1,3 +1,15 @@
...
+
+- Fixed an intermittent `OSError` on Windows when DLLs are loaded or unloaded
+  concurrently during library discovery (for example when importing conda-forge
+  OpenCV). Windows module enumeration now uses a snapshot-first approach with
+  graceful per-module fallbacks.
+  https://github.com/joblib/threadpoolctl/issues/217
+
+- Added Windows integration tests for library paths longer than `MAX_PATH`.
+  https://github.com/joblib/threadpoolctl/pull/189
+
```
ogrisel: Let's keep the changelog focussed on user facing changes.

```suggestion
```

tests/test_threadpoolctl.py
```diff
@@ -792,3 +799,69 @@ def test_custom_controller():
         assert mylib_controller.num_threads == 1
 
     assert ThreadpoolController().info() == original_info
+
+
+def test_threadpool_controller_repeated_init():
```
dugalh, line 805: It's probably worthwhile running this as is, but in my testing I have only seen its code fail when it is run after ``import cv2`` in a Windows conda-forge environment.

threadpoolctl.py
```diff
@@ -1039,80 +1043,237 @@ def _find_libraries_with_dyld(self):
...
         PROCESS_QUERY_INFORMATION = 0x0400
         PROCESS_VM_READ = 0x0010
 
-        LIST_LIBRARIES_ALL = 0x03
-
         ps_api = self._get_windll("Psapi")
         kernel_32 = self._get_windll("kernel32")
+        self._setup_windows_module_apis(ps_api, kernel_32)
 
         h_process = kernel_32.OpenProcess(
```
dugalh: You could use [``kernel_32.GetCurrentProcess()``](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentprocess) here.  It returns a pseudo-handle that doesn't need to be closed.

threadpoolctl.py
```diff
@@ -1039,80 +1043,237 @@ def _find_libraries_with_dyld(self):
...
+
+            if modules is not None:
+                for h_module, snapshot_path in modules:
+                    filepath = self._resolve_module_filepath(
+                        ps_api,
+                        h_process,
+                        h_module,
+                        snapshot_path,
+                        max_path,
+                        path_buf,
```
dugalh: I don't think this is necessary here.  In my testing, I only see ``_snapshot_loaded_modules()`` succeed and return paths when they are all complete and shorter than ``MAX_PATH``.  If any path of a loaded DLL is longer than ``MAX_PATH`` (actually 255), ``CreateToolhelp32Snapshot()`` fails.

threadpoolctl.py
```diff
@@ -1039,80 +1043,237 @@ def _find_libraries_with_dyld(self):
...
+        h_process,
+        h_module,
+        snapshot_path,
+        max_path,
+        path_buf,
+    ):
+        """Return the full path for a module, or None if it should be skipped."""
+        from ctypes.wintypes import MAX_PATH
+
+        if ps_api.GetModuleFileNameExW(h_process, h_module, path_buf, max_path):
```
dugalh, line 1350: I suggest using [``kernel_32.GetModuleFileNameW()``](https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulefilenamew).  It is ~10x faster.

tests/utils.py
dugalh: I'm not too sure about the changes in ``utils.py``.  I think you are better qualified to make a call on those.

tests/test_threadpoolctl.py
ogrisel: Alright, I think we should update the CI to install opencv and the test to import cv2 and skip it if not available.

tests/utils.py
```diff
@@ -86,3 +104,73 @@ def select(info, **kwargs):
...
+        return sorted(
+            libopenblas_paths,
+            key=lambda path: (
+                0
+                if "libscipy_openblas" in os.path.basename(path).lower()
+                else 1 if "libopenblas" in os.path.basename(path).lower() else 2
+            ),
+        )[0]
+
+    from threadpoolctl import ThreadpoolController
```
ogrisel: Move that import to the top.

tests/utils.py
```diff
@@ -86,3 +104,73 @@ def select(info, **kwargs):
...
+        return path
+    if path.startswith("\\\\"):
+        return "\\\\?\\UNC\\" + path[2:]
+    return "\\\\?\\" + path
+
+
+def copy_and_load_dll(src_dll, destination):
+    """Copy a DLL to destination and load it in the current process."""
+    shutil.copy2(src_dll, to_extended_windows_path(destination))
+    ctypes.CDLL(to_extended_windows_path(destination))
```
ogrisel: Let's remove such trivial helper and expand the code were needed in the tests.

tests/utils.py
```diff
@@ -17,6 +20,17 @@
...
+    libopenblas_patterns.append(
+        os.path.join(numpy_site_packages, "numpy.libs", "libscipy_openblas*.dll")
+    )
+    if sys.platform == "win32":
+        libopenblas_patterns.append(
+            os.path.join(sys.prefix, "Library", "bin", "openblas*.dll")
+        )
+        libopenblas_patterns.append(
+            os.path.join(sys.prefix, "Library", "bin", "libopenblas*.dll")
+        )
```
ogrisel:
```suggestion
    if sys.platform == "win32":
    libopenblas_patterns.append(
        os.path.join(sys.prefix, "Library", "bin", "openblas*.dll")
    )
    libopenblas_patterns.append(
        os.path.join(sys.prefix, "Library", "bin", "libopenblas*.dll")
    )
```

tests/utils.py
```diff
@@ -86,3 +104,73 @@ def select(info, **kwargs):
...
+    path = os.path.abspath(str(path))
+    if path.startswith("\\\\?\\"):
+        path = path[4:]
+        if path.startswith("UNC\\"):
+            path = "\\\\" + path[4:]
+    return os.path.normcase(os.path.normpath(path))
+
+
+LONG_PATH_OPENBLAS_DLL = "libopenblas_long_path_test.dll"
+TRUNCATED_PATH_OPENBLAS_DLL = "libopenblas_path_too_long.dll"
```
ogrisel: Let's remove those constants and use inline string literals in the tests instead.

tests/test_threadpoolctl.py
```diff
@@ -792,3 +800,73 @@ def test_custom_controller():
...
+    """Stress-test repeated library discovery.
+
+    Non-regression test for a Windows-specific problem where DLLs loaded or
+    unloaded concurrently during discovery could raise an OSError. The failure
+    was originally reproduced after importing OpenCV on Windows conda-forge;
+    see https://github.com/joblib/threadpoolctl/issues/217
+    """
+    pytest.importorskip("cv2")
+    import cv2  # noqa: F401
```
ogrisel:
```suggestion
```

ogrisel (commented):
Code style review.

threadpoolctl.py
ogrisel: Done in 9b325f3: snapshot paths are used directly via `_snapshot_module_filepath()`; `GetModuleFileNameExW` is only used in the `EnumProcessModulesEx` fallback.

threadpoolctl.py
ogrisel: Done in 9b325f3.

threadpoolctl.py
ogrisel: Done in 9b325f3: fallback enumeration tries `GetModuleFileNameW` first, then `GetModuleFileNameExW` for long paths.

tests/test_threadpoolctl.py
ogrisel: Done in b2efac3: the test imports `cv2` via `pytest.importorskip` and OpenCV is installed in CI on the Windows jobs.

tests/utils.py
ogrisel: Done in 4aa9633.

tests/utils.py
ogrisel: Done in 4aa9633: removed `copy_and_load_dll`, `normalize_windows_path`, and the DLL name constants; logic is inlined in the Windows tests.

tests/utils.py
ogrisel: Done in 4aa9633.

tests/utils.py
ogrisel: Done in 4aa9633.

tests/test_threadpoolctl.py
ogrisel: Done in 4aa9633: `pytest.importorskip("cv2")` is enough to load OpenCV.

threadpoolctl.py
```diff
@@ -1039,79 +1043,239 @@ def _find_libraries_with_dyld(self):
...
         This function is expected to work on windows system only.
-        This code is adapted from code by Philipp Hagemeister @phihag available
-        at https://stackoverflow.com/questions/17474574
+
+        Module discovery uses a snapshot-first strategy: when
+        ``CreateToolhelp32Snapshot`` succeeds, ``szExePath`` values are already
+        complete and shorter than ``MAX_PATH``. When the snapshot fails (for
+        example because a loaded DLL lives on a long path), enumeration falls
+        back to ``EnumProcessModulesEx`` and resolves paths with
+        ``GetModuleFileNameExW`` using a larger buffer.
```
```suggestion
        Module discovery uses a snapshot-first strategy: when
        ``CreateToolhelp32Snapshot`` succeeds, ``szExePath`` values are already
        complete and shorter than ``MAX_PATH``. When the snapshot fails (for
        example because a loaded DLL lives on a long path), enumeration falls
        back to ``EnumProcessModulesEx`` and resolves paths with
        ``GetModuleFileNameExW`` using a larger buffer. The snapshot based
        approach is more robust in case of concurrent DLL loading/unloading,
        however it cannot handle long paths, hence the need for the fallback.
```

ogrisel (commented):
I think this LGTM. The licensing issue is still there but will be addressed in a follow-up PR to decouple the concerns.

cc @jeremiedbb @tomMoral.

ogrisel:
It seems that clang conda-forge packaging has changed because it can no longer find its openmp but it's unrelated to this PR.

ogrisel:
I synced with master and everything is green thanks to @lesteve's recent CI fixes.

Shall we merge?

lesteve:
Codex `/review` tells me the following. I am afraid that I haven't tried to understand yet whether it makes sense or not, very sorry about that :sweat: 

### [P1] Bound snapshot retries so long paths reach the fallback

**File:** `threadpoolctl.py:1312-1313`

When `CreateToolhelp32Snapshot` persistently returns `ERROR_BAD_LENGTH`—notably in the long-DLL-path scenario this fallback is intended to handle—this unconditional retry loops forever, so `_find_libraries_on_windows` never catches an
`OSError` and never invokes `EnumProcessModulesEx`.

Limit the retries and raise after exhaustion so discovery can use the long-path-capable fallback.

lesteve:
I can reproduce the original on my Windows VM so I'll give it a try as a sanity check.

ogrisel:
Here is the reply from agent to your agent:

The Codex diagnosis is incorrect, but a retry cap would still be a reasonable hardening change.

`ERROR_BAD_LENGTH` (24) is the documented *transient* race when the module list changes while `CreateToolhelp32Snapshot` is taken. MSDN's own wording for `TH32CS_SNAPMODULE` / `TH32CS_SNAPMODULE32` is to retry until it succeeds. The `while True` loop is that guidance, not a forgotten bound.

Long DLL paths are a different failure mode: `MODULEENTRY32W.szExePath` is limited to `MAX_PATH`. If the snapshot succeeds, we get a truncated path and skip that library with a warning. The `EnumProcessModulesEx` fallback is only used when snapshot creation raises `OSError`, not when `szExePath` is truncated. `ERROR_BAD_LENGTH` is not the long-path error.

`test_windows_library_path_longer_than_max_path` already exercises that case. If long paths caused a persistent `ERROR_BAD_LENGTH` spin, that test would hang; it does not.

A bounded retry (say 8-16 attempts, no sleep) then `raise OSError(...)` would still be good library hygiene, since `ThreadpoolController()` sits on a hot path and an unbounded loop could hang if `GetLastError` were ever sticky or mis-attributed. I would treat that as optional hardening, not a merge blocker for #217.

A related but separate gap: if the snapshot *succeeds* with a truncated `szExePath`, we currently skip the library instead of resolving that one handle with `GetModuleFileNameExW`. That is independent of the infinite-loop claim.

ogrisel:
@copilot resolve the merge conflicts in this pull request

ogrisel:
I synced with `main`. The good news is that on newer Python versions, `dllist` is used instead of the ctypes based win32 codepath (as of the merge of #248) and the non-regression tests still pass \o/

So this means that the ugly extra complexity introduced in that PR to get those non-regression test to also pass on older CPython version will go away in a few years once we stop support CPython < 3.14.

I think this PR deserves to be merged, at least to introduce the non-regression and stress tests. If we don't want to introduce the extra complexity, we could revert it and skip the new tests on older Python versions.

WDYT @lesteve?

ogrisel:
Actually, my above comment is wrong. The custom win32 code branch is always used, even on newer CPython versions.

Let's check if this is needed or not.

ogrisel:
Let's see if the new tests fail with dlllist on windows.

ogrisel:
This is great: the dllist method passes all the non-regression stress tests on Windows with recent versions of Python. This means that once we drop older Python support we will be able to significantly trim the windows specific code path, including the extra complexity introduced in this PR.

ogrisel (commented):
LGTM. The CI is really stable, even when running the stress test on more cases.

Issue #220, opened 2026-07-09
Thread-local limits for OpenBLAS with pthreads


itamarst:
## The problem

OpenBLAS with pthreads is widely used: a billion downloads of NumPy from PyPI a month; 400 million for SciPy. Not all are Linux, but most are. Not all of them use BLAS, of course, but some do. It adds up!

The API limit for threads is process-wide. This is annoying, but not the worst part.

The worst part is that it has a process-wide shared pool. Consider a CPU with 32 cores, and a Python thread pool with 4 threads, each of which wants to do some OpenBLAS work. A user whose code needs to run with a variety of BLAS libraries (e.g. sklearn committer) sets a limit `32 // 4 == 8` cores. For many setups this is fine; for OpenBLAS with pthreads, it means 24 cores sit idle.

## Potential solutions

* Change this upstream in OpenBLAS.
* Get PyPI packages to stop shipping pthreads version and instead use OpenMP version. So far all OpenMP libraries I've seen don't have process-wide shared thread pool, so this will make things better in most cases.
* Turning https://github.com/itamarst/openblas-tod/ into a real project, and having its own thread-local-scoped limiting API.
* Rethink scope limiting in `threadpoolctl` in this case.

To expand on the last point—

## Maybe `threadpoolctl` should do something different for OpenBLAS with pthreads (or more broadly APIs affecting a process-wide thread pool)

A user might set three different settings:

1. All cores. "All" means... all cores if it's a top-level process, or whatever share of cores were assigned to this process if it's a worker in a process pool.
2. Run inline/sequentially, without a thread pool, typically set via 1 or 0.
3. A number in between.

Typically the first case will be used when there are no Python threads, so it should be exactly the right number. The second case will be used when there is a Python thread per-core, so again exactly the right number.

The only problems occur, then, in the third case. As we saw in the example above, you may end up with a bunch of cores sitting idle. So maybe in the third case it's worth changing the user limit to something better.

To make this concrete, imagine 16 cores, and a Python thread pool that calls BLAS but also does other work, perhaps in a single thread, perhaps in additional thread pools (Python thread pools, OpenMP, etc, who can say...).

| Python thread pool size | Naive OpenBLAS limit| Better OpenBLAS limit |
|------------------------------------|---------------------------------|----------------------------------------------|
| 2   | 8     | All cores might be better, 16 instead of 8, but risks oversaturation |
| 3   | 5     | All cores might be better, 16 instead of 5, but risks oversaturation |
[...]

Issue #214, opened 2026-06-23
API to introspect semantic meaning of setting a thread pool limit


itamarst:
The scope (thread-local / process-wide) is covered by #211.

This issue covers introspecting if setting the number of threads:

1. Is a per-thread limit for the current thread only. This is (probably) always the case if the API scope is current thread only.
2. Assuming the API's scope affects the whole process, is it setting a per-thread limit for the whole process (what MKL does)...
3. ...or, is it changing the size of a process-wide global thread pool (OpenBLAS with pthreads).

#211 has some discussion of how one might approach this.

It seems, apriori, to be more tricker than scope, so perhaps it shouldn't be a user-facing piece of information, but rather just something run in CI to ensure any future docs/annotations match the real semantics.

PR #221, opened 2026-07-10
Add private Windows ETW thread spawn tracer prototype

ogrisel:
Context:

- https://github.com/joblib/threadpoolctl/issues/220#issuecomment-4927122863
- https://github.com/joblib/threadpoolctl/issues/214

The purpose of this cursor-driven PR is to study the feasibility of using only ctypes to trace thread start events on Windows from a Python program.

Introduce threadpoolctl._thread_tracer as an experimental private package with a ctypes-based WindowsThreadSpawnTracer that counts kernel Thread Start events for a target PID via the NT Kernel Logger.

Includes pure-Python event payload parsing, mocked ETW lifecycle tests that run on all platforms, and keeps the tracer out of the public threadpoolctl API.

I do not have a local windows setup, so I will let the agent leverage the Windows CI of this repo to iterate.


threadpoolctl/_thread_tracer/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+def windows_etw_admin_available():
+    """Return whether the current process can start kernel ETW sessions."""
+    if not windows_tracer_available():
+        return False
+    try:
+        import ctypes
+
+        return bool(ctypes.windll.shell32.IsUserAnAdmin())
+    except (AttributeError, OSError):
+        return False
```
ogrisel: Let's remove those trivial functions and move that logic directly inside the skipif predicates in the tests instead.

_thread_tracer/_windows_etw.py
```diff
@@ -0,0 +1,531 @@
...
+
+
+class EVENT_TRACE_LOGFILE(ct.Structure):
+    pass
+
+
+if sys.platform == "win32":
+    EVENT_RECORD_CALLBACK = ct.WINFUNCTYPE(None, ct.POINTER(EVENT_RECORD))
+else:
+    EVENT_RECORD_CALLBACK = ct.CFUNCTYPE(None, ct.POINTER(EVENT_RECORD))
```
ogrisel: Let's simplify this. This code is not expected to be imported on non-Windows platforms. Let's not imply otherwise.

ogrisel:
Note: I plan to leave this PR in draft mode for now but the results of the integration tests run on the windows CI seem to indicate that it works as expected. Unfortunately, I don't think it's possible to do the same for macOS and Linux without relying on a helper compiled extension or on a helper system tracing tool that might require extra privileges.

PR #222, opened 2026-07-10
FIX only warn about libomp/libiomp conflict on Linux

ogrisel:
Tentative fix Windows `pylatest_conda_forge_mkl` CI failure.

The conda-forge MKL environment loads both `libiomp` and `libomp`, triggering `RuntimeWarning` on every `ThreadpoolController` init. pytest runs with `-W error`, so tests fail.

The incompatibility between Intel and LLVM OpenMP runtimes is documented as a Linux-specific issue. This change skips the warning on non-Linux platforms.

Note: I am not sure if this is the right fix, but it's interesting to check if the crash is indeed Linux specific or not.

The alternative would be to change the dependencies in the CI config so that the Windows CI jobs that need MKL would not load clang's OpenMP runtime.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,8 +2,7 @@
 ===========
 
 - Only warn about simultaneous `libomp` and `libiomp` usage on Linux, where the
-  incompatibility is known to cause crashes. This avoids spurious test failures on
-  Windows CI jobs that load both libraries via conda-forge MKL stacks.
+  incompatibility is known to cause crashes.
 
 3.6.0 (2025-03-13)
 ==================
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,10 @@
+3.7.0 (TBD)
+===========
+
+- Only warn about simultaneous `libomp` and `libiomp` usage on Linux, where the
+  incompatibility is known to cause crashes. This avoids spurious test failures on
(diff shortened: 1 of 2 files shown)


ogrisel:
Maybe we should also check that the libiomp + libomp crash on linux is still happening with the latest versions of those libraries. It could be the case that they have become more compatible overtime and that raising a warning is no longer needed on any platform.

CHANGES.md
```diff
@@ -1,3 +1,10 @@
+3.7.0 (TBD)
+===========
+
+- Only warn about simultaneous `libomp` and `libiomp` usage on Linux, where the
+  incompatibility is known to cause crashes. This avoids spurious test failures on
+  Windows CI jobs that load both libraries via conda-forge MKL stacks.
```
```suggestion
  incompatibility is known to cause crashes.
```

ogrisel:
Let's merge this to get a clean CI ASAP and I will follow-up with more testing to check if warning on Linux is still needed or not.

itamarst (approved):
Insofar as I can review things: this seems reasonable to me.

ogrisel:
For the record, we can still reproduce the crash on linux using the reproducers in https://github.com/ogrisel/threadpoolctl/pull/14. So let's keep the warning on linux.

PR #224, opened 2026-07-13
Thread limit introspection, part 2: Empirical observation in CI

### LINKED ISSUE #214: API to introspect semantic meaning of setting a thread pool limit

itamarst:
Fixes #214 

In theory one could, say, distribute another package in PyPI/conda-forge of compiled OpenMP code so users can run this locally. Or I guess ... pick some part of e.g. scikit-learn that is unlikely to change but that seems brittle too. So this could be better, and maybe sometime it's worth doing that.

But at least it should run in CI and give us useful info, so it's a good first step.


itamarst:
Looks like occasionally the output is erroneously "not sure", but mostly gets everything right.

ogrisel:
Is this PR still needed if we merge #228?

itamarst:
I think so - the technique I use for testing in #228 is Linux-specific (`strace`). There's apparently dtrace on macOS but that's restricted by System Integrity Protection, not sure what you'd do on Windows, etc.. So this PR is a nice OS-agnostic approach that is still informative.

ogrisel:
Ok, I'll add it to my review backlog.

For Windows, feel free to take over or inspiration from this vibecoded proof of concept: https://github.com/joblib/threadpoolctl/pull/221.

tests/empirical_scope_observation.py
```diff
@@ -0,0 +1,123 @@
...
+    # assume process-wide shared thread pool. If the number is close to double
+    # the number of Python threads, we'll assume a thread pool per thread.
+    if max_num_threads < num_python_threads * 1.25:
+        scope = "process-wide shared thread pool"
+    elif max_num_threads > num_python_threads * 1.75:
+        scope = "thread pool per thread"
+    else:
+        scope = "not sure"
+
+    print(f"== Observed behavior: {which} ==")
```
ogrisel, line 116: Let's make the CI log easier to interpret for people who did not read the source code of this script:

```suggestion
    print(f"== Observed behavior: {which} ==")
    print(
        f"10x{cpu_count()} Python threads each running a parallel "
        "workload with a 2 thread limit."
    )
```

tests/empirical_scope_observation.py
```diff
@@ -0,0 +1,123 @@
...
+from pprint import pprint
+from threading import Thread
+from typing import Callable
+
+import threadpoolctl
+import psutil
+
+
+def start_counting_threads() -> Callable[[], int]:
+    num_threads = 0
```
ogrisel: Nit: I would rather call this `max_num_threads`.

tests/empirical_scope_observation.py
```diff
@@ -0,0 +1,123 @@
...
+
+def blas(num_python_threads):
+    try:
+        import numpy as np
+    except ImportError:
+        print("BLAS not available")
+        return False
+
+    set_limits()
+    A = np.ones((10_000_000,))
```
ogrisel: Not sure if it's needed but I would rather exercise BLAS level 3 that is more arithmetic intensive to make sure that multithreading is always expected to be beneficial and therefore actually used by all BLAS runtimes for this workload.

```suggestion
    A = np.ones((1024, 1024))
```

tests/empirical_scope_observation.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def blas_math(_):
+        A.dot(A)
+
+    with ThreadPoolExecutor(num_python_threads) as pool:
+        for _ in pool.map(blas_math, range(400)):
+            pass
```
ogrisel: Maybe we could try something like the following to have a better control on the workload duration and make sure that psutil has the time time to inspect meaningful thread usage data for this process from the OS runtime.

```suggestion
        t0 = time()
        while time() - t0 < 5:
            for _ in pool.map(blas_math, range(40)):
                pass
```

tests/empirical_scope_observation.py
```diff
@@ -0,0 +1,123 @@
...
+        return False
+
+    set_limits()
+
+    def run_openmp(_):
+        check_openmp_num_threads(1000)
+
+    with ThreadPoolExecutor(num_python_threads) as pool:
+        for _ in pool.map(run_openmp, range(400)):
+            pass
```
ogrisel: We could also use a time based outer loop here.

ogrisel (approved):
Some improvement suggestions but otherwise LGTM.

ogrisel:
Note: I noticed that `threadpoolctl.threadpool_info` fails to detect opennblas (or any other BLAS) on some Windows builds. The conda package list still reports openblas as installed and therefore probably used by numpy. This is not a problem for that PR but this is probably a new bug.

EDIT: I opened https://github.com/joblib/threadpoolctl/issues/231 to track that issue.

itamarst:
OK I've addressed all review comments.

ogrisel:
Merged! Thanks again @itamarst.

PR #227, opened 2026-08-20
DBG empty commit to trigger CI in branch

ogrisel:
Check if I can reproduce macOS specific CI build issues.


lesteve:
This will be fixed by #237.

PR #228, opened 2026-08-31
Use thread-local limiting APIs when possible, and fix deadlock

### LINKED ISSUE #216: Use thread-local APIs when possible

### LINKED ISSUE #229: Potential deadlock when loading shared libraries, specifically mkl with libiomp
When loading shared libraries, the mechanism on Linux to enumerate shared libraries is the POSIX `dl_iterate_phdr` API. This takes a callback that is called on every loaded library, and in threadpoolctl this calls back into Python.

The callback then loads the library with `dlopen()`, and gets some info, which among other thing ends up calling `mkl_get_num_threads()`.

This becomes a problem in the case of mkl with libiomp, perhaps because it does its own `dlopen()` (loading "libarcher.so") when `mkl_get_num_threads()` is called leading to some sort of reentrancy.

I haven't figured out _exactly_ how this triggers a deadlock, it requires multiple threads and happens consistently when `tests/_limit_blas.py` is called in #228 (including if you run that file on `master`, it's not about any changes I made in the PR).

The solution, to be added to #228, is to minimize the work done in the Python callback so there's no reentrancy into mkl/dl/anything else.

itamarst:
Fixes #216 
Fixes #229 

I figured out a way to measure thread-creation that works on Linux, at least. Or that's the hope, anyway. (`ltrace` would be much more robust but I can't get it to work on Conda binaries).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,23 @@
 - Only warn about simultaneous `libomp` and `libiomp` usage on Linux, where the
   incompatibility is known to cause crashes.
 
+- Fixed a deadlock triggered by getting or setting MKL's number of threads from
+  parallel threads when using MKL with libiomp (Intel threading) on Linux.
+
+- Going forward, setting the number of threads will only have a thread-local
+  impact if feasible (for example, at minimum the underlying library must
+  support this option, and many don't.)
+
+- For MKL, setting the number of threads is now thread-local, i.e. limiting the
+  number of threads won't impact MKL's thread pool size when using MKL in other
+  Python threads.
+
+- For OpenBLAS compiled with OpenMP on Linux and macOS, setting the number of
+  threads is now thread-local, i.e. won't impact OpenBLAS thread pool size in
+  other Python threads. On Windows behavior is likely process-wide, but this may
+  depend on how OpenBLAS was compiled with OpenMP.
+
+
 3.6.0 (2025-03-13)
 ==================
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,20 +12,23 @@
 
 - Fixed a deadlock triggered by getting or setting MKL's number of threads from
   parallel threads when using MKL with libiomp (Intel threading) on Linux.
+  https://github.com/joblib/threadpoolctl/pull/228
 
 - Going forward, setting the number of threads will only have a thread-local
   impact if feasible (for example, at minimum the underlying library must
   support this option, and many don't.)
+  https://github.com/joblib/threadpoolctl/pull/228
 
 - For MKL, setting the number of threads is now thread-local, i.e. limiting the
   number of threads won't impact MKL's thread pool size when using MKL in other
   Python threads.
+  https://github.com/joblib/threadpoolctl/pull/228
 
 - For OpenBLAS compiled with OpenMP on Linux and macOS, setting the number of
   threads is now thread-local, i.e. won't impact OpenBLAS thread pool size in
   other Python threads. On Windows behavior is likely process-wide, but this may
   depend on how OpenBLAS was compiled with OpenMP.
-
+  https://github.com/joblib/threadpoolctl/pull/228
 
 3.6.0 (2025-03-13)
 ==================
--- a/tests/test_api_introspection.py
+++ b/tests/test_api_introspection.py
@@ -60,9 +60,6 @@ def test_determine_thread_limit_scope_processwide(default: int) -> None:
     assert _determine_thread_limit_scope(api.get, api.set) == "process"
 
 
-@pytest.mark.skipif(
-    sys.platform != "linux", reason="Non-Linux OpenMP might be different"
-)
 @pytest.mark.parametrize(
     ["select_filter", "expected_thread_limit_scope", "extra_check"],
     [
@@ -77,7 +74,12 @@ def test_determine_thread_limit_scope_processwide(default: int) -> None:
             # pthreads here.
             lambda lib: lib.threading_layer == "pthreads",
         ),
-        ({"user_api": "openmp"}, "current_thread", lambda _lib: True),
+        (
+            {"user_api": "openmp"},
+            "current_thread",
+            # Windows OpenMP is process-wide:
+            lambda _lib: sys.platform in ("linux", "darwin"),
+        ),
     ],
 )
 def test_api_scope(
@@ -94,9 +96,11 @@ def test_api_scope(
     if not controller.lib_controllers:
         pytest.skip(f"{select_filter} controller not found")
 
-    for lib in controller.lib_controllers:
-        if not extra_check(lib):
-            pytest.skip("extra check returned false")
+    libs = [lib for lib in controller.lib_controllers if extra_check(lib)]
+    if not libs:
+        pytest.skip("No libraries matched the requirements")
+
+    for lib in libs:
         assert (
             _determine_thread_limit_scope(lib.get_num_threads, lib.set_num_threads)
             == expected_thread_limit_scope
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -874,7 +874,7 @@ def test_setting_limit_on_thread_local_blas_api_is_actually_thread_local(
             ],
             stderr=subprocess.STDOUT,
         ).splitlines():
-            if b" clone3(" in line and b"CLONE_THREAD" in line:
+            if b"clone3(" in line and b"CLONE_THREAD" in line:
                 result += 1
         print(limit, result)
         return result
--- a/tests/test_threadpoolctl.py
+++ b/tests/test_threadpoolctl.py
@@ -6,10 +6,12 @@ import pytest
 import re
 import subprocess
 import sys
+from typing import Callable
 
 from threadpoolctl import threadpool_limits, threadpool_info
-from threadpoolctl import ThreadpoolController
+from threadpoolctl import LibController, ThreadpoolController
 from threadpoolctl import _ALL_PREFIXES, _ALL_USER_APIS
+from threadpoolctl import _determine_thread_limit_scope
 
 from .utils import cython_extensions_compiled
 from .utils import check_nested_prange_blas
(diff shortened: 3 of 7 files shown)


itamarst:
Not sure what's up with openmp suddenly failing to compile on macOS. Seems unrelated to this PR.

tests/test_api_introspection.py
```diff
@@ -77,7 +74,12 @@ def test_determine_thread_limit_scope_processwide(default: int) -> None:
             # pthreads here.
             lambda lib: lib.threading_layer == "pthreads",
         ),
-        ({"user_api": "openmp"}, "current_thread", lambda _lib: True),
+        (
+            {"user_api": "openmp"},
+            "current_thread",
+            # Windows OpenMP is process-wide:
```
ogrisel, line 80: I asked an LLM to know if this was a bug or not. Apparently, this is not a bug but a consequence of a being an implementation of an older version of the spec:

```suggestion
            # Windows OpenMP is process-wide as it claims to implement
            # the OpenMP 2.0 spec. This would be a violation of the
            # OpenMP 3.0 spec which mandates per-task
            # Internal Control Variables such as `nthreads-var`.
            # https://learn.microsoft.com/en-us/cpp/build/reference/openmp-enable-openmp-2-0-support?view=msvc-180
```

For information, recent MSVC can be configured to build with the `-openmp:llvm` flag to leverage libomp instead: https://devblogs.microsoft.com/cppblog/improved-openmp-support-for-cpp-in-visual-studio/

I don't know if there is an easy way to snif libomp specific symbols on such MSVC generated binary files.

tests/test_threadpoolctl.py
```diff
@@ -795,3 +798,88 @@ def test_custom_controller():
...
+        pytest.skip(f"{select_filter} controller not found")
+
+    libs = [lib for lib in controller.lib_controllers if extra_check(lib)]
+    if not libs:
+        pytest.skip("No libraries matched the requirements")
+
+    return libs
+
+
+def test_setting_limit_on_thread_local_blas_api_is_reported_as_thread_local(
```
ogrisel, line 846: Should we add a similar test for the case(s) where we expect a process-wide scope?

threadpoolctl.py
```diff
@@ -552,7 +573,9 @@ def get_num_threads(self):
         return get_func()
 
     def set_num_threads(self, num_threads):
-        set_func = getattr(self.dynlib, "MKL_Set_Num_Threads", lambda num_threads: None)
+        set_func = getattr(
+            self.dynlib, "MKL_Set_Num_Threads_Local", lambda num_threads: None
+        )
```
ogrisel, line 583: We need to document that behavior change in a changelog entry.

threadpoolctl.py
```diff
@@ -289,13 +293,30 @@ def set_additional_attributes(self):
...
+        # The OpenBLAS limiting API is process-wide, and we want current thread
+        # limit if possible. When OpenBLAS is backed by OpenMP, using the
+        # OpenMP API allows for current thread limiting when OpenMP has that
+        # behavior. That is the case for libgomp, libomp, and libiomp, what you
+        # would find on Linux or macOS. On Windows the Visual C++ OpenMP API is
+        # process-wide, unfortunately. Also worth knowing that before v0.3.34,
+        # the OpenBLAS limiting API is broken when using OpenMP threading:
+        # https://github.com/OpenMathLib/OpenBLAS/issues/5806
+        if self.threading_layer == "openmp" and sys.platform in ("linux", "darwin"):
+            symbol = "omp_set_num_threads"
```
ogrisel, line 321: We need to document that behavior change in a changelog entry.

threadpoolctl.py
```diff
@@ -1120,16 +1143,18 @@ def _find_libraries_with_dl_iterate_phdr(self):
...
             return []
 
+        filepaths = []
+
         # Callback function for `dl_iterate_phdr` which is called for every
-        # library loaded in the current process until it returns 1.
+        # library loaded in the current process until it returns 1. To minimize
+        # the potential for deadlocks (see #228), this code should not do
+        # anything that might result in reentrancy into the library, the dl
+        # system, or anything else.
```
ogrisel, line 1157: We should also document this potential fix in a changelog entry.

ogrisel (commented):
Thanks very much @itamarst for the PR. Overall, this LGTM but there are a few points to address:

tests/test_api_introspection.py
itamarst: I imagine there's no reason _not_ to use omp_set_num_threads on Windows, it might just not be thread-local (but it won't be worse than openblas_set_num_threads). So as a first pass I will:

* Change the code to run on Windows too.
* Still leave the test disabled on Windows.
* Open an issue to see if detection can be done.
* ~Open an issue with OpenBLAS on Conda-Forge to use this flag, with the presumption I will verify that it doesn't already.~

tests/test_threadpoolctl.py
itamarst: Mm. The ideal is everything is thread-local, so if it behaves thread-locally without us noticing that's fine. This test is more for "did we successfully opt-in to the desired behavior".

itamarst: So, probably not necessary.

tests/test_api_introspection.py
itamarst: Ah, sounds like that VS flag is experimental for now, so maybe I won't suggest it.

itamarst:
Not really sure what's going on with the new compiler failures...

itamarst:
Dug into compiler failures a bit: https://github.com/conda-forge/python-feedstock/pull/914#issuecomment-5514167090

itamarst:
There's a new PR in conda-forge/python-feedstock to fix the compilation issue, so hopefully by tomorrow it should be working again, will just need to re-run.

ogrisel:
I retriggered a build and the compiler problems went away but `pylatest_flexiblas` has the `test_setting_limit_on_thread_local_blas_api_is_actually_thread_local[thread_local_blas_libs1]` that fails with:

```python
        # _limit_blas runs BLAS operations in 2 Python threads, so by changing the
        # BLAS limit from 1 to 4 we expect an extra 2 * (4 - 1) == 6 threads.
        extra_threads = num_threads_created(4) - num_threads_created(1)
>       assert extra_threads == 6
E       assert 0 == 6
```

on top of the unrelated macOS OpenMP failures.

tests/test_threadpoolctl.py
```diff
@@ -795,3 +798,90 @@ def test_custom_controller():
...
+                and parse_version(lib.version) >= parse_version("0.3.34")
+            ),
+        ),
+        (
+            {"internal_api": "mkl"},
+            lambda _lib: True,
+        ),
+    ]
+)
+def thread_local_blas_libs(request) -> list[LibController]:
```
ogrisel: I think we should let pytest display a more informative string representation for each parametrisation. At the moment we get the following in the pytest output:

```
tests/test_threadpoolctl.py::test_setting_limit_on_thread_local_blas_api_is_actually_thread_local[thread_local_blas_libs0] SKIPPED [ 98%]
tests/test_threadpoolctl.py::test_setting_limit_on_thread_local_blas_api_is_actually_thread_local[thread_local_blas_libs1] FAILED [100%]
```

I think we should pass a callable or a string list to the `ids=` kwarg of the `pytest.fixture`:

```python
ids=["openblas-openmp", "mkl"]
```

CHANGES.md
```diff
@@ -10,6 +10,23 @@
...
+  support this option, and many don't.)
+
+- For MKL, setting the number of threads is now thread-local, i.e. limiting the
+  number of threads won't impact MKL's thread pool size when using MKL in other
+  Python threads.
+
+- For OpenBLAS compiled with OpenMP on Linux and macOS, setting the number of
+  threads is now thread-local, i.e. won't impact OpenBLAS thread pool size in
+  other Python threads. On Windows behavior is likely process-wide, but this may
+  depend on how OpenBLAS was compiled with OpenMP.
```
ogrisel, line 30: Let's add the link to the PR for each entry.

ogrisel (commented):
More feedback on the new iteration of this PR.

tests/test_api_introspection.py
ogrisel: I think it's still worth expanding the inline comment to explain that the current windows OpenMP runtime implement semantics from an older version of the OpenMP spec but that it is hopefully expected to change in the future, at which point we might want to update this test to make sure it passes on all platforms.

tests/test_threadpoolctl.py
```diff
@@ -795,3 +798,90 @@ def test_custom_controller():
...
+            stderr=subprocess.STDOUT,
+        ).splitlines():
+            if b"clone3(" in line and b"CLONE_THREAD" in line:
+                result += 1
+        return result
+
+    # _limit_blas runs BLAS operations in 2 Python threads, so by changing the
+    # BLAS limit from 1 to 4 we expect an extra 2 * (4 - 1) == 6 threads.
+    extra_threads = num_threads_created(4) - num_threads_created(1)
+    assert extra_threads == 6
```
ogrisel: Hopefully pytest will display more informative traceback information in case of failure:

```suggestion
    nmc_1 = num_threads_created(1)
    nmc_4 = num_threads_created(4)
    assert nmc_4 - nmc_1 == 6
```

itamarst:
> I retriggered a build and the compiler problems went away but `pylatest_flexiblas` has the `test_setting_limit_on_thread_local_blas_api_is_actually_thread_local[thread_local_blas_libs1]` that fails with:
> 
> ```python
>         # _limit_blas runs BLAS operations in 2 Python threads, so by changing the
>         # BLAS limit from 1 to 4 we expect an extra 2 * (4 - 1) == 6 threads.
>         extra_threads = num_threads_created(4) - num_threads_created(1)
> >       assert extra_threads == 6
> E       assert 0 == 6
> ```
> 
> on top of the unrelated macOS OpenMP failures.

I've seen occasional results (in CI only) where the number of threads is unexpectedly when number of threads was increased. My current theory is that maybe OpenMP caps number of threads to number of cores, so I'll see if I can figure out hardware used in that run.

tests/test_threadpoolctl.py
itamarst: That helps yes.

itamarst:
Oh I wonder if that's just because it's _flexiblas_. So setting the number of threads is using the flexiblas API which doesn't use the `_local()` MKL variant. I'll open an issue for flexiblas to look at separately since I don't want that to block this PR.

ogrisel (approved):
LGTM!

PR #234, opened 2026-09-03
Document using limiting APIs for Python thread pools

### LINKED ISSUE #215: API for setting limits for use by Python thread pools

itamarst:
Fixes #215

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,10 @@ Fine control of the underlying thread-pool size can be useful in
 workloads that involve nested parallelism so as to mitigate
 oversubscription issues.
 
+Note that "limiting the number of threads" in practice involves a variety of
+potential semantics depending on the underlying third-party library; see the
+section on [semantics](#semantics) below.
+
 ## Installation
 
 - For users, install the last published version from PyPI:
@@ -458,11 +462,98 @@ https://github.com/xianyi/OpenBLAS/issues/2985).
   the underlying library. For more details see
   https://github.com/joblib/threadpoolctl/issues/208
 
-  For example, if you're using OpenMP with libgomp (gcc) or libomp (clang), the
+  For example, if you're using OpenMP with libgomp (gcc) or libomp (clang**, the
   setting is thread-local and sets how many OpenMP threads will be started in
   the current thread. On the other hand, with OpenBLAS with pthreads backend or
   on Windows, the setting is process-wide and impacts the size of a process-wide
   thread pool shared across all threads in the process.
+
+## Semantics of thread limiting <a name="semantics">
+
+Setting the number of threads may seem like a simple operation, but in practice
+it can do quite different things depending on the underlying third-party library
+being limited.
+
+### Kind of thread pool
+
+* In some libraries, there is a shared process-wide thread pool. Setting the
+  number of threads changes the size of this shared pool.
+* In other libraries, there is a thread pool per calling thread. So each Python
+  thread gets its own personal thread pool from the third-party library.
+
+To give a concrete example: if you have 10 cores, and you're using OpenBLAS with
+the `pthreads` backend on Linux, by default there is a single 10-worker pool of
+threads created by OpenBLAS. If you start 10 Python threads, each calling BLAS
+routines, all those Python threads will feed into that single 10-thread pool.
+In total, you will have 20 threads running (10 Python, 10 OpenBLAS).
+
+On the other hand, if you use MKL, each Python thread gets its own individual
+pool of worker threads from MKL, so if each Python threads calls a BLAS routine
+in MKL, you will get 10×10 = 100 MKL threads!
+
+### Scope of limiting
+
+If you have a third-party library that creates a thread pool per calling thread,
+there is another question about what limiting the number of threads means: which
+threads are affected by the limiting API?
+
+* **All threads in the process:** In this case, setting the limits changes it
+  for all threads in the process. So if you limit from the main Python thread,
+  all other Python threads will have the same limit.
+* **Current thread only:** Only the current thread's thread pool size will be
+  limited.
+
+MKL for example has two APIs, covering both cases,
+[`mkl_set_num_threads()`](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2026-0/mkl-set-num-threads.html)
+and [`mkl_set_num_threads_local()`](https://www.intel.com/content/www/us/en/docs/onemkl/developer-reference-c/2026-0/mkl-set-num-threads-local.html) respectively.
+
+### `threadpoolctl`'s policy for thread limiting
+
+When there is a choice between different APIs, `threadpoolctl` will prefer APIs
+that:
+
+1. Have a per-thread worker pool (so each Python thread gets its own pool from
+   the underlying third-party library)
+2. Only affect the current thread.
+
+Current libraries where `threadpoolctl` explicitly makes this choice are:
+
+* MKL, for all threading backends.
+* OpenBLAS (v0.3.34 or later) when using the OpenMP backend, on Linux and macOS.
+
+### Checking the behavior of your installed libraries
+
+When you run `python -m threadpoolctl` it will include the scope of the API
+limit in the `"thread_limit_scope"` field, with the value of `"process"`
+indicating the API affects the whole process and `"current_thread"` indicating a
+per-thread thread pool. Information about the kind of limiting in the former
+case (shared process-wide pool, or per-thread pool) is not included.
+
+Here is example output for OpenBLAS with pthreads:
+
+```shell-session
+$ python -m threadpoolctl -i numpy
+[
+  {
+    "user_api": "blas",
+    "internal_api": "openblas",
+    "num_threads": 12,
+    "prefix": "libscipy_openblas",
+    "version": "0.3.33.112.0",
+    "threading_layer": "pthreads",
+    "architecture": "Haswell",
+    "thread_limit_scope": "process"
+  }
+]
+```
+
+You can also use another script to empirically determine both the kind and scope of the API:
+
+```shell-session
+
+```
+
+You can also call this script with `openmp` instead of `blas`.
 
 ## Maintainers
 
--- a/README.md
+++ b/README.md
@@ -187,17 +187,11 @@ however not act on libraries loaded after the instantiation of the
 ...     a_squared = a @ a
 ```
 
-This should only be used for APIs you expect to be called from the main thread
-of a process, without the use of any Python thread pools.
-
 ### Restricting the Limits to the Scope of a Function
 
 `threadpool_limits` and `ThreadpoolController` can also be used as decorators to set
 the maximum number of threads used by the supported libraries at a function level. The
 decorators are accessible through their `wrap` method.
-
-This should only be used for functions you expect to be called from the main
-thread of a process, without the use of any Python thread pools.
 
 ```python
 >>> from threadpoolctl import ThreadpoolController, threadpool_limits
@@ -221,10 +215,10 @@ This section covers APIs to use when you will be using Python thread pools to pa
 ### Setting the Maximum Size of Thread-Pools, When Python Thread Pools Are Used
 
 Limiting thread pool size in controlled libraries requires a two-step process.
-Importantly, each Python thread must call a method to limit controlled libraries
-in that thread. With Python's `concurrent.futures.ThreadPoolExecutor`, you can
-do so by passing in an initializer function that will get called on thread
-startup.
+**Importantly, each Python worker thread must also call a method to limit
+controlled libraries in that thread.** With Python's
+`concurrent.futures.ThreadPoolExecutor`, you can do so by passing in an
+initializer function that will get called on thread startup.
 
 ```python
 from threadpoolctl import threadpool_limits
@@ -234,11 +228,9 @@ with threadpool_limits(limits=1) as limiter:
     # Make sure each Python worker thread also calls threadpool_limits(). If
     # you're using another thread pool class, you will need to do so some other
     # way.
-    with ThreadPoolExecutor(
-            4,
-            initializer=lambda: threadpool_limits(limits=1)
-    ) as pool:
+    with ThreadPoolExecutor(4, initializer=lambda: threadpool_limits(limits=1)) as pool:
         # ... run some BLAS-using code in the thread pool ...
+        pool.map(somefunc, someargs)
 ```
 
 To prevent loading shared libraries repeatedly, you can reuse a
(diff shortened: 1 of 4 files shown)


itamarst:
I think once this merged maybe this could be released?

README.md
```diff
@@ -205,6 +208,120 @@ decorators are accessible through their `wrap` method:
...
+        pool.map(somefunc, someargs)
+finally:
+    limiter.restore_original_limits()
+
+```
+
+### Switching Back And Forth Between Main Thread and Python Threads
+
+Unfortunately not all controlled libraries providing limiting APIs that are
+thread-specific. Limiting some libraries' thread pool sizes can therefore impact
```
ogrisel, line 289: I think it would be helpful to maintain a separate markdown file with a reference assesment of the known semantics of common libraries that we consolidated by running the helpers from #213 and #214 on our CI.

Here is my current understanding:

- OpenMP (including BLAS runtimes configured with an OpenMP threading layer):
  - libgomp (GCC) and libomp (LLVM): thread-local limits and thread-pools as per the OpenMP 3.0+ spec
  - msvc: process-wide limits and shared thread-pool (as per the OpenMP 2.0 spec)
- OpenBLAS with the pthreads threading layer: process-wide limits and shared thread-pool
- OpenBLAS with the windows threading layer: process-wide limits and shared thread-pool
- MKL (all threading layers?): thread-local limits and thread-pools

Alternatively, we could have a short README section that explains how to use the CLI and the Python script that allows the user to dynamically inspect the effective semantics as we do on our CI.

README.md
```diff
@@ -205,6 +208,120 @@ decorators are accessible through their `wrap` method:
...
+    def limit_then_do_work(*args, **kwargs):
+        # Set a limit on OpenMP in the current thread:
+        CONTROLLER.limit(limits=1)
+        # Do the actual work:
+        return do_real_work_with_openmp(*args, **kwargs)
+
+    results = POOL.map(limit_then_do_work, args)
+
+
+# 2. Run some work directly in main thread, using OpenMP.
```
ogrisel: It feels weird to state "with OpenMP" when the point is to disable parallel execution.


```suggestion
# 2. Run some work serially in main thread (OpenMP effectively disabled).
```

README.md
```diff
@@ -205,6 +208,120 @@ decorators are accessible through their `wrap` method:
...
+uses these libraries in Python threads and running it in the main thread a bit
+more complex: you need to set the limits each time you switch back and forth.
+
+Let's say your computer has 4 cores, and you're using some OpenMP API.
+
+```python
+POOL = ThreadPoolExecutor(4)
+CONTROLLER = ThreadpoolController()
+
+# 1. Run some work in a Python thread pool, which then runs in OpenMP.
```
ogrisel:
```suggestion
# 1. Run some work in a Python thread pool (OpenMP effectively disabled).
```

README.md
```diff
@@ -205,6 +208,120 @@ decorators are accessible through their `wrap` method:
...
+
+    results = POOL.map(limit_then_do_work, args)
+
+
+# 2. Run some work directly in main thread, using OpenMP.
+with CONTROLLER.limit(limits=1):
+    results2 = do_more_work_with_openmp(results)
+
+
+# 3. Do more work in a Python thread pool, this time with more parallelism:
```
ogrisel:
```suggestion
# 3. Nest some OpenMP parallelism under Python-level parallelism:
```

ogrisel (commented):
A first pass of feedback. I think it would be helpful if someone else who did not review all your past PRs could have a fresh look at this. Maybe @cakedev0, @lesteve and/or @jeremiedbb?

README.md
itamarst: I went with the second approach, let me know what you think.

README.md
```diff
@@ -205,6 +212,120 @@ decorators are accessible through their `wrap` method:
...
+
+```
+
+### Switching Back And Forth Between Main Thread and Python Threads
+
+Unfortunately not all controlled libraries providing limiting APIs that are
+thread-specific. Limiting some libraries' thread pool sizes can therefore impact
+the whole process. This makes switching back and forth between running code that
+uses these libraries in Python threads and running it in the main thread a bit
+more complex: you need to set the limits each time you switch back and forth.
```
ogrisel, line 295: I think it's worth linking to the semantics section from here to explain why this extra complexity is needed.

```suggestion
Unfortunately not all controlled libraries providing limiting APIs that are
thread-specific, as detailed in the section on [semantics](#semantics) below.
Limiting some libraries' thread pool sizes can therefore impact the whole
process. This makes switching back and forth between running code that
uses these libraries in Python threads and running it in the main thread a bit
more complex: you need to set the limits each time you switch back and forth.
```

ogrisel (approved):
This is great. LGTM.

ogrisel:
BTW, the CI should now be green if you sync this PR's branch with the master branch.

README.md
```diff
@@ -205,6 +212,120 @@ decorators are accessible through their `wrap` method:
...
+    # Make sure each Python worker thread also calls threadpool_limits(). If
+    # you're using another thread pool class, you will need to do so some other
+    # way.
+    with ThreadPoolExecutor(4, initializer=lambda: threadpool_limits(limits=1)) as pool:
+        # ... run some BLAS-using code in the thread pool ...
+        pool.map(somefunc, someargs)
+```
+
+To prevent loading shared libraries repeatedly, you can reuse a
+`ThreadpoolController` object:
```
cakedev0: I don't really understand what "loading shared libraries repeatedly" means and costs.

README.md
```diff
@@ -205,6 +212,120 @@ decorators are accessible through their `wrap` method:
...
+        CONTROLLER.limit(limits=1)
+        # Do the actual work:
+        return do_real_work_with_openmp(*args, **kwargs)
+
+    results = POOL.map(limit_then_do_work, args)
+
+
+# 2. Run some work serially in main thread (OpenMP effectively disabled).
+with CONTROLLER.limit(limits=1):
+    results2 = do_more_work_with_openmp(results)
```
cakedev0: Why would you want to run some serial OMP workload in the main thread?

It's more likely that I want to let this step run with the default settings I'd say. It's a bit hard to understand this, when step 1 and 3 both use 4 threads.

Maybe:

```suggestion
# 2. Run some work with 4-threads OpenMP parallelism:
with CONTROLLER.limit(limits=4):
    results2 = do_more_work_with_openmp(results)
```

README.md
```diff
@@ -205,6 +212,120 @@ decorators are accessible through their `wrap` method:
...
+
+
+# 3. Nest some OpenMP parallelism under Python-level parallelism:
+with CONTROLLER.limit(limits=2) as limiter:
+
+    def limit_then_do_work2(*args, **kwargs):
+        limiter.limit(limits=2)
```
cakedev0: `limiter` here is a `_ThreadpoolLimiter` instance, which has no `.limit()` method — calling this raises `AttributeError`. Did you mean `CONTROLLER.limit(limits=2)`, as in the examples above?

README.md
```diff
@@ -340,12 +462,107 @@ https://github.com/xianyi/OpenBLAS/issues/2985).
   the underlying library. For more details see
   https://github.com/joblib/threadpoolctl/issues/208
 
-  For example, if you're using OpenMP with libgomp (gcc) or libomp (clang), the
+  For example, if you're using OpenMP with libgomp (gcc) or libomp (clang**, the
```
cakedev0: why the closing parenthesis got replaced with `**`?

conftest.py
```diff
@@ -1 +1,5 @@
-collect_ignore = ["tests/_openmp_test_helper", "tests/_limit_blas"]
+collect_ignore = [
+    "tests/empirical_scope_observation",
```
cakedev0: This entry is missing the `.py` extension, so it wont match `tests/empirical_scope_observation.py` — the ignore has no effect (harmless here only because the file also wouldnt match pytests default collection pattern).

pyproject.toml
```diff
@@ -27,3 +27,6 @@ homepage = "https://github.com/joblib/threadpoolctl"
 line-length = 88
 target_version = ['py39', 'py310', 'py311', 'py312', 'py313']
 preview = true
+
+[tool.ruff]
+line-length = 88
```
cakedev0: Missing trailing newline at end of file.

README.md
```diff
@@ -340,12 +462,107 @@ https://github.com/xianyi/OpenBLAS/issues/2985).
...
   setting is thread-local and sets how many OpenMP threads will be started in
   the current thread. On the other hand, with OpenBLAS with pthreads backend or
   on Windows, the setting is process-wide and impacts the size of a process-wide
   thread pool shared across all threads in the process.
 
+## Semantics of thread limiting <a name="semantics">
```
cakedev0: This `<a name="semantics">` anchor is never closed with `</a>`, unlike other named anchors in the doc.

cakedev0: This break the md preview in VS-code, what do you think of:

```suggestion
## <div id="semantics"> Semantics of thread limiting </div>
```

README.md
```diff
@@ -152,6 +143,22 @@ The state of these libraries is also accessible through the object oriented API:
 True
 ```
 
+## Usage when not using Python threads: Restricting Controlled Library Thread Pool Sizes
+
+There a two scenarios in which you might want to use `threadpoolctl`; each
```
cakedev0:
```suggestion
There are two scenarios in which you might want to use `threadpoolctl`; each
```

README.md
```diff
@@ -205,6 +212,120 @@ decorators are accessible through their `wrap` method:
...
+    # ... run some BLAS-using code in the thread pool ...
+    pool.map(somefunc, someargs)
+```
+
+You can also operate without a context manager:
+
+```python
+from threadpoolctl import ThreadpoolController
+
+CONTROLLER = ThreadpoolController():
```
cakedev0:
```suggestion
CONTROLLER = ThreadpoolController()
```

README.md
```diff
@@ -205,6 +212,120 @@ decorators are accessible through their `wrap` method:
...
+try:
+    limiter = controller.limit()
```
cakedev0:
```suggestion
    limiter = CONTROLLER.limit()
```

README.md
```diff
@@ -205,6 +212,120 @@ decorators are accessible through their `wrap` method:
...
+    with ThreadPoolExecutor(
+            4, initializer=lambda: controller.limit(limits=1)) as pool:
```
cakedev0:
```suggestion
            4, initializer=lambda: CONTROLLER.limit(limits=1)) as pool:
```

cakedev0 (commented):
Some small findings here and there. Plus two genuine, possibly naive, questions. Otherwise, I found it relatively clear and easy to understand.

README.md
itamarst: Stupid bug in my editor's auto-complete.

itamarst:
I believe I've addressed all the review comments, thanks for the detailed feedback.

README.md
```diff
@@ -241,7 +241,8 @@ with threadpool_limits():
         pool.map(somefunc, someargs)
 ```
 
-To prevent loading shared libraries repeatedly, you can reuse a
+Whenever `threadpool_limits` is called, it needs to do some work (inspecting and getting access to third-party shared libraries) that can take some time.
+To prevent the performance cost of doing this work every time, you can reuse a
```
cakedev0: :+1:

cakedev0 (approved):
LGTM, based on the diff since my last review.

Thank you for the nice documentation effort! :smile:

ogrisel:
Thanks again! I will follow-up with a release soon.

cakedev0:
If you also plan a release of loky, can we please merge https://github.com/joblib/loky/pull/651 before? :pray:

lesteve:
I think it's only a release for threadpoolctl for now, but I haven't forgotten your loky PR :wink:.

itamarst:
Note I filed another issue that is a release blocker for threaded support, hoping to open a PR today.

PR #237, opened 2026-09-06
CI Fix compilation issue with OpenMP inside conda env

### LINKED ISSUE #232: CI is failing on macOS due to some OpenMP compilation issues
E.g. https://github.com/joblib/threadpoolctl/actions/runs/33778674267

lesteve:
Fix https://github.com/joblib/threadpoolctl/issues/232

Here is a summary of my investigation:
- there is a bug on macOS for nested login shells https://github.com/conda/conda/issues/13500, the PATH is in the wrong order so system clang is picked instead of the `clang` symlink in the conda env. This cause the error about `clang` no knowing what to do with `-fopenmp` 
- minimally activated conda compilers is part of the picture https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/8795. Now what happens is that `CC` is not set when you install `compilers`. So `clang` is the compiler rather than the full conda compiler executable i.e. `x86_64-conda-<something>`. Before `CC` being set would hide the macOS bug from previous bullet point
- there is a number of combinatorial stuff that I tried before converging to something that works and doesn't seem to brittle/dodgy. I would rather avoid trying more things :wink:.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -239,7 +239,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          ./continuous_integration/install.sh
+          source continuous_integration/install.sh
 
       # - name: Open SSH debug session
       #   uses: owenthereal/action-upterm@v1
@@ -251,7 +251,7 @@ jobs:
 
       - name: Test library
         run: |
-          ./continuous_integration/run_tests.sh
+          source continuous_integration/run_tests.sh
 
       - name: Upload test results
         uses: actions/upload-artifact@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -241,6 +241,13 @@ jobs:
         run: |
           bash -el continuous_integration/install.sh
 
+      - name: Open SSH debug session
+        uses: owenthereal/action-upterm@v1
+        if: ${{ always() }}
+        with:
+          limit-access-to-actor: true
+          wait-timeout-minutes: 60
+
       - name: Test library
         run: |
           bash -el continuous_integration/run_tests.sh
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -241,14 +241,6 @@ jobs:
         run: |
           ./continuous_integration/install.sh
 
-      # - name: Open SSH debug session
-      #   uses: owenthereal/action-upterm@v1
-      #   if: ${{ failure() }}
-      #   with:
-      #     limit-access-to-actor: true
-      #     wait-timeout-minutes: 60
-      #     shell: bash --noprofile --norc
-
       - name: Test library
         run: |
           ./continuous_integration/run_tests.sh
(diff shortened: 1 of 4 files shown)


lesteve:
Really? [build log](https://github.com/joblib/threadpoolctl/actions/runs/34045958702/job/101520985655?pr=237)
```
continuous_integration/run_tests.sh: line 24:   742 Segmentation fault         PYTHONPATH=. python tests/empirical_scope_observation.py blas
Error: Process completed with exit code 139.
```

No idea where that was coming from but maybe when I was using `source continuous_integration/<script>.sh`, I haven't been able to reproduce with just `./continuous_integration/<script>.sh`. I don't think it's worth investigating, but feel free to investigate if you don't agree with me :wink:.

.github/workflows/test.yml
```diff
@@ -16,7 +16,7 @@ on:
 # Cancel in-progress workflows when pushing
 # a new commit on the same branch
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
lesteve, line 19: This wasn't cancelling the build before when pushing a new commit, I took it from scikit-learn

.github/workflows/test.yml
```diff
@@ -214,7 +214,7 @@ jobs:
       run:
         # Need to use this shell to get conda working properly.
         # See https://github.com/marketplace/actions/setup-miniconda#important
-        shell: ${{ matrix.os == 'windows-latest' && 'cmd /C CALL {0}' || 'bash -el {0}' }}
+        shell: 'bash -el {0}'
```
lesteve, line 217: This made my life easier to always use bash, probably an historical thing to use CMD on Windows, not 100% sure

.github/workflows/test.yml
```diff
@@ -239,11 +239,11 @@ jobs:
 
       - name: Install dependencies
         run: |
-          bash -el continuous_integration/install.sh
+          ./continuous_integration/install.sh
```
lesteve, line 242: not using `bash -l` avoids the nested login shell bug in macOS where system clang is picked first rather than conda env symlink

continuous_integration/install.sh
```diff
@@ -46,6 +46,8 @@ make_conda() {
         TO_INSTALL="$TO_INSTALL python-gil"
     fi
 
+    # Need to source conda.sh before first conda command
+    source "$CONDA/etc/profile.d/conda.sh"
```
lesteve, line 50: sourcing the conda.sh file is the simpler way to make `conda` command work on all OSes inside a script. Previously we were relying on using login shell `bash -l` which kind of work because setup-miniconda adds the equivalent to `.profile` or something sourced at login. This was kind of brittle because of the macOS PATH bug within nested login shell. Also I have seen cases where it behaves weirdly on Windows, I don't remember the details.

lesteve (commented):
Comments about the changes to help review

ogrisel (approved):
Thank you very much for the investigation and fix @lesteve! Let's merge.

Issue #225, opened 2026-08-06
threadpool_info() before os.fork() causes SIGABRT on child exit (Python 3.13+)


bornexmachina:
## `threadpool_info()` before `os.fork()` causes SIGABRT on child exit (Python 3.13+)

**Versions:** threadpoolctl 3.6.0, Python 3.14.5, libffi 3.4.x (Linux)

Hi all, 
I have updated my code library from Python3.9 to 3.14. I have noticed that stopping my Python service produced core dumps. My service has a parent which forks multiple children which use scikit-learn.

I could reproduce a minimal running code with a coredump. Sorry, I am not sure, whether it is threadpoolctl or CPython issue. Thus, I open the issue with CPython as well. 

Below you will find the structured output summarized by our glorious AI overlords.

Thanks! 

### Problem

Calling `threadpoolctl.threadpool_info()` in a parent process before
`os.fork()`, then letting the forked child exit via `sys.exit()`, reliably
crashes the child with `SIGABRT` inside libffi's allocator:

```
abort
dlfree.cold (libffi.so.8)
CThunkObject_dealloc (_ctypes)
... (Py_Finalize / module teardown) ...
```

### Cause

`threadpoolctl` imports `ctypes.util.find_library`. On Python 3.13+,
`ctypes.util` implements Linux lookups via `dl_iterate_phdr()` using a
module-level `ctypes.CFUNCTYPE` closure (`ctypes.util._info_callback`) that
lives for the process lifetime. If this closure is created in a parent
*before* `fork()`, the forked child inherits it via copy-on-write. libffi's
private closure allocator keeps global free-list state that diverges
between parent and child after fork; when the child later frees the
inherited closure at interpreter shutdown, the allocator aborts on
corrupted metadata.

### Minimal repro

```python
import os, sys, signal, time
import threadpoolctl

threadpoolctl.threadpool_info()  # creates ctypes.util._info_callback

def child():
    signal.signal(signal.SIGTERM, lambda s, f: sys.exit(0))
    threadpoolctl.threadpool_info()
    time.sleep(30)

pid = os.fork()
if pid == 0:
    child()
else:
    time.sleep(3)
    os.kill(pid, signal.SIGTERM)
    os.waitpid(pid, 0)
```

`python3.14 repro.py` → `Aborted (core dumped)`.

Does **not** crash if `threadpoolctl.threadpool_info()` is only called
after the fork (i.e., not in the parent).

bornexmachina:
[...]

PR #242, opened 2026-09-08
Avoid ctypes callbacks when listing libraries on Linux.

ogrisel:
Workaround for #225 on some system:

Importing ctypes.util or using CFUNCTYPE with dl_iterate_phdr creates libffi closures that can abort after fork on some Linux libffi builds. Enumerate libraries from /proc/self/maps and load libc via CDLL(None).

Disclaimer: this PR was vibecoded although I did ask the agent to use a container to check that it actually workaround the problem on AlmaLinux 9.8 (Olive Jaguar) with libffi-3.4.2-8.el9.

Still, I find the extra complexity in threadpoolctl not justified. I would rather wait for upstream ctypes to fix the problem or wait for all impacted linux distributions to upgrade to a newer libffi.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,13 @@
 3.7.0 (TBD)
 ===========
 
+- On Linux, loaded libraries are now enumerated from ``/proc/self/maps``
+  instead of a ``ctypes`` callback into ``dl_iterate_phdr``, and libc is
+  loaded with ``ctypes.CDLL(None)`` instead of ``ctypes.util.find_library``.
+  Both previous paths created libffi closures that can abort after ``os.fork()``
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -41,6 +41,12 @@
 
   - On Linux, start using /proc/self/maps for listing shared libraries.
 
+- Avoid importing ``ctypes.util`` on Linux (and load libc with
+  ``ctypes.CDLL(None)``) so ``threadpool_info()`` does not create libffi
+  closures that can abort after ``os.fork()`` on some libffi builds.
+  https://github.com/joblib/threadpoolctl/issues/225
```
```suggestion
```

ogrisel:
I think the extra code complexity is fine, let's merge.

PR #249, opened 2026-09-15
PERF Go faster by caching library detection and loading

### LINKED ISSUE #246: Speed up usage of threadpoolctl
Initial benchmarking with more extensive use of threadpoolctl in scikit-learn suggest that the overhead of setting limits can be considerable. While there may be ways to improve that within joblib and/or scikit-learn, having threadpoolctl itself be faster would help all users.

Some ideas:

* Microoptimizations.
* Caching the `ThreadpoolController`, and invalidating it when imports change.
* Caching individual `CDLL`. This has the nice property that one can reasonably assume they never get unloaded, so you don't need invalidation logic.
* ... maybe more as I look at profiling results and think ...

itamarst:
Fixes #246

See benchmarks below


threadpoolctl.py
```diff
@@ -264,6 +267,137 @@ def _get_symbol(self, name):
...
+class _CachingCDLL:
+    """Wrap a ``CDLL``, caching attributes."""
+
+    _cdll: ctypes.CDLL
+    _cache: dict[str, ctypes._CFuncPtr | None] = field(default_factory=dict)
+
+    def __getattr__(self, symbol: str) -> ctypes._CFuncPtr | None:
+        result = self._cache.get(symbol, _MISSING)
+        if result is _MISSING:
+            result = getattr(self._cdll, symbol, None)
```
ogrisel, line 284: Out of curiosity, do you have quantitative evidence that caching attribute looks of `ctypes.CDLL` instances is useful? I would not have expected them to be expensive.

itamarst: It showed up in the profile, pretty sure, which is what motivated this. When I get to the final PR I will give some numbers without this enabled, and remove it turns out to be unnecessary.

itamarst:
Will re-open when it's ready for review.

itamarst:
Benchmarks, `python benchmarks/bench_context_manager_overhead.py --n-calls 50_000 --import numpy` with 3.14t.

* Master branch: 0.125 +/-0.291 ms
* This PR: 0.0705 +/-0.0688 ms.

threadpoolctl.py
itamarst: I ended up removing it, since it didn't help in the final version. I still am pretty sure it's theoretically useful insofar as it showed up in profiles, but in practice the caching of method results I added later supersedes it, since it caches at a higher level.

tests/test_cache.py
```diff
@@ -0,0 +1,44 @@
...
+    # ``cache_method_on_dynlib()``, which is currently the case.
+    controller = ThreadpoolController()
+    libs = controller.select(user_api="blas").lib_controllers
+    if not libs:
+        pytest.skip("No libraries loaded")
+
+    assert libs[0].get_version() is libs[0].get_version()
+
+    # Access underlying, uncached get_version():
+    assert libs[0].get_version.__wrapped__(libs[0]) is not libs[0].get_version()
```
ogrisel: Isn't that assertion brittle in the sense that it could fail because of string interning?

Maybe a monkeypatch / mock would be more robust way to assess how many times the uncached underlying method is called when calling the cached public method wrapper several times in a row.

threadpoolctl.py
```diff
@@ -289,6 +292,159 @@ def _get_symbol(self, name):
...
+        return None
+
+    def create_controller(
+        self, filepath: str, parent: ThreadpoolController
+    ) -> LibController | None:
+        """
+        Create the associated controller, if there is one, relying on cached
+        info for the given ``filepath``.
+        """
+        result = self._controller_cache.get(filepath, _MISSING)
```
ogrisel, line 338: This means that we should be able to drop the need for a singleton factory in sklearn, right?

https://github.com/scikit-learn/scikit-learn/blob/d059607efd6de710a059563e33cea5df9b915d10/sklearn/utils/parallel.py#L187

Or is it still worth caching the `ThreadpoolController` instance?

ogrisel (commented):
I confirm I also oberseve a factor of 2+ reduction in the `threadpool_limits` call (with the `sklearn` module imported) on my macOS host: from 0.5 ms per call on `master` vs 0.2 ms per call on this branch.

threadpoolctl.py
ogrisel: Listing the libraries is still not cheap so it's worth keeping a single `ThreadpoolController` instance for scikit-learn, unless we also merge #254.

tests/test_cache.py
itamarst: Yeah, probably. Will do that.

threadpoolctl.py
itamarst: Yeah, it's still slow on Linux in common setups (any Python before 3.15, to begin with).

itamarst:
OK should be ready for review again.

ogrisel (approved):
LGTM!

Let's wait a bit for @jeremiedbb to give his own feedback.

PR #250, opened 2026-09-16
PERF Run faster on Linux

### LINKED ISSUE #248: The new Linux /proc/self/maps shared library listing is very slow
This change has made `threadpool_limits()` run almost 3× as slow.

(This is separate from #246 because this issue is very Linux specific; that issue will cover more generic optimizations.)

itamarst:
Fixes #248 

Benchmarked with `python benchmarks/bench_context_manager_overhead.py --n-calls 10_000 --import numpy` on 3.14 free-threaded.

* 3.7.0: `0.223 +/-0.256 ms`
* This PR: `0.127 +/-0.251 ms`

Switch to `dllist()` is even faster, but not a good idea before Python 3.15.


tests/test_threadpoolctl.py
```diff
@@ -1131,19 +1131,24 @@ def test_controller_parallelism_no_deadlocks():
     if sys.platform != "linux" or not hasattr(ctypes.PyDLL(None), "backtrace"):
```
ogrisel:
```suggestion
    if sys.platform != "linux":
```

tests/test_threadpoolctl.py
```diff
@@ -1131,19 +1131,24 @@ def test_controller_parallelism_no_deadlocks():
...
-    # Internally, backtrace() calls dl_iterate_phdr which can result in
-    # deadlocks if threadpoolctl is also using dl_iterate_phdr.
-    backtrace_gil = ctypes.PyDLL(None).backtrace
-    backtrace_gil.argtypes = [ctypes.c_void_p, ctypes.c_int]
-    backtrace_nogil = ctypes.CDLL(None).backtrace
-    backtrace_nogil.argtypes = [ctypes.c_void_p, ctypes.c_int]
+    done = []
 
     def create_controllers():
         buf = (ctypes.c_void_p * 20)()
```
ogrisel:
```suggestion
```

tests/test_threadpoolctl.py
```diff
@@ -1131,19 +1131,24 @@ def test_controller_parallelism_no_deadlocks():
...
         for _ in range(100):
+            # May use dl_iterate_phdr() on Linux:
             limiter = threadpool_limits()
-            backtrace_gil(buf, 20)
-            backtrace_nogil(buf, 20)
+            # dlopen():
+            try:
+                # Should be available in most Linux, and importantly isn't
+                # loaded by default into Python:
+                dll = ctypes.CDLL("libncurses.so.6")
```
ogrisel: Wouldn't it be better to run this in a subprocess to make this test stronger by ensuring that dll loading is guaranteed to happen instead of assuming that this particular dynlib has not been already silently loaded elsewhere when running the test suite?

We could increase concurrent DLL loading pressure by also attempting to import Python modules with more typical dynlibs handled by controllers such as the one that are created when importing numpy's BLAS, the OpenMP linked into the Cython extensions built for testing purpose.

threadpoolctl.py
```diff
@@ -56,6 +56,37 @@
...
+    # Make sure /proc is mounted:
+    and os.path.exists("/proc/self")
+    # If dllist() is available (3.14+), written in C so no risk of GC part way
+    # (3.15+), and there is no GIL, no need to use /proc, since dllist() is
+    # faster. It's possible dllist() might work in GIL builds too but see
+    # https://github.com/python/cpython/issues/157573. So we should consider
+    # enabling it on GIL Python too once threadpoolctl supports 3.15.
+    and not (
+        sys.version_info[:2] >= (3, 15)
+        and not getattr(sys, "_is_gil_enabled", lambda: True)()
```
ogrisel: Let's add a linux free-threading CPython 3.15 config to the CI to exercise this condition.

ogrisel (commented):
Thanks for the fix. LGTM but I think we could improve the testing harness as follows:

itamarst:
Thank you, I'll address these today hopefully.

threadpoolctl.py
itamarst: I realized that doing this at import isn't great, because importing extension modules that don't support free-threading can enable the GIL. So I'll move this inside the function, alas.

itamarst:
I'm getting close to just giving up on 3.15 environments for now; either one would switch to installing Python via GitHub Actions, or just wait a month or three until conda-forge is caught up.

itamarst:
Yeah I'm giving up on this for now. Will revert those changes.

threadpoolctl.py
itamarst: I tried, but it's a real pain to get things going with release candidates. I suggest waiting until it's no longer pre-release and doing it in #251.

ogrisel:
Did you run the tests locally on free-threading Python 3.15?


EDIT: I just did and everything looks good.

.github/workflows/test.yml
```diff
@@ -221,6 +221,26 @@ jobs:
             CC_OUTER_LOOP: "gcc"
             CC_INNER_LOOP: "gcc"
 
+          # Python 3.14, free-threaded
+          - name: py314_freethreaded_pip_openblas
+            os: ubuntu-latest
+            PACKAGER: "conda-forge"
+            BLAS: "openblas"
+            PYTHON_VERSION: "3.14"
```
ogrisel: This way we won't have to change the CI config once Python 3.15 is out and we don't really care about free-threaded users on older Python versions since the free-threaded stack is not yet widely adopted.

```suggestion
          # Free-threaded python
          - name: pylatest_freethreaded_pip_openblas
            os: ubuntu-latest
            PACKAGER: "conda-forge"
            BLAS: "openblas"
            PYTHON_VERSION: "*"
```

ogrisel (approved):
LGTM. Will merge and follow-up with a dedicated PR to add 3.15 CI config and official support in the metadata.

itamarst:
Thank you!

Issue #238, opened 2026-09-07
Remove Python 3.9 testing and mentions


lesteve:
Python 3.9 security support has ended in October 2025
```
❯ git grep -P '3.*9'
.github/workflows/test.yml:          - name: py39_pip
.github/workflows/test.yml:            PYTHON_VERSION: "3.9"
.github/workflows/test.yml:          - name: py39_conda_homebrew_libomp
.github/workflows/test.yml:            PYTHON_VERSION: "3.9"
.github/workflows/test.yml:          - name: py39_ubuntu_atlas_gcc_gcc
.github/workflows/test.yml:            PYTHON_VERSION: "3.9"
.github/workflows/test.yml:          - name: py39_ubuntu_openblas_gcc_gcc
.github/workflows/test.yml:            PYTHON_VERSION: "3.9"
.github/workflows/test.yml:          # Linux + Python 3.9 and homogeneous runtime nesting.
.github/workflows/test.yml:          - name: py39_conda_openblas_clang_clang
.github/workflows/test.yml:            PYTHON_VERSION: "3.9"
CHANGES.md:3.5.0 (2024-04-29)
continuous_integration/install.sh:    # https://github.com/scikit-learn/scikit-learn/pull/13934
pyproject.toml:requires-python = ">=3.9"
pyproject.toml:    "Programming Language :: Python :: 3.9",
pyproject.toml:target_version = ['py39', 'py310', 'py311', 'py312', 'py313']
```

cc @DeaMariaLeon

itamarst:
Wait another month or two and you can drop 3.10 at the same time.

PR #255, opened 2026-09-21
Remove Python 3.9 support

DeaMariaLeon:
Addresses #238 

It looks like `black` pinned version 25.1.0 doesn't support Python 3.14, ~~so I didn't add it.~~
https://github.com/psf/black/blob/25.1.0/src/black/mode.py


pyproject.toml
```diff
@@ -27,7 +26,7 @@ homepage = "https://github.com/joblib/threadpoolctl"
 
 [tool.black]
 line-length = 88
-target_version = ['py39', 'py310', 'py311', 'py312', 'py313']
+target_version = ['py310', 'py311', 'py312', 'py313']
```
ogrisel: We can probably add 'py314' and maybe even 'py315' to the mix.

ogrisel (approved):
Thanks @DeaMariaLeon. LGTM once the following is addressed.

pyproject.toml
DeaMariaLeon: To add 'py314' I had to change black's pinned version and change a text format in `threadpoolctl.py` to make the CI happy.

ogrisel (approved):

ogrisel:
Thanks for the PR. Maybe we could switch to ruff if you are interested in doing a follow-up contrib.

DeaMariaLeon:
Yes, sure. Thanks.

PR #256, opened 2026-09-21
Document just the faster idiom for Python thread pools

### LINKED ISSUE #253: Modify docs to emphasize faster pattern for Python thread pools
Having determined that the time spent loading libraries can add up in real-world, it's worth only documenting the faster pattern for Python thread pools:

1. Create one `ThreadpoolController`
2. Re-use it in the threads

It may also be worth documenting while I'm at it the limits of having a single `ThreadpoolController`: not noticing the existence of new libraries, usually due to later imports.

itamarst:
Fixes #253 

Once this and #249 are merged, should probably do a new release.


ogrisel (approved):
LGTM. Thanks.

